PR #1, opened 2015-02-13
Fixed error on Rust master - no drop flag requires feature

vhbit:


vhbit:
Note: it builds buf for test pass it also requires https://github.com/cgaebel/iobuf/pull/8 to land

PR #3, opened 2015-02-16
Remove unnecessary Send impl

fhartwig:
`ByteBuf` does not need to explicitly implement `Send`, since all its components are `Send` (because `MemRef` already implements `Send`).


carllerche:
Looks legit. This was carry over from pre MemRef

carllerche:
Merged: 7e02c29e85c07180e5a6e8a4af8202d41a3d1f68

Issue #4, opened 2015-02-22
Removing the RingBufReader and RingBufWriter types


fhartwig:
This is really a question more than an issue:
Could we remove the RingBuf Reader and Writer newtypes (and implement `Buf` and `MutBuf` directly on `RingBuf`)? I don't actually understand what purpose they have, apart from disambiguating method names from `Buf` and `MutBuf` (but that could be done in other ways).
If we could get rid of them, that would massively improve the ergonomics of `RingBuf`, at least for my use case. There are two problems that I have with them:
* There are about a dozen instances in my code where I have to write something along the lines of `let r = buf.reader(); let s = r.bytes();` where I would like to write `let s = buf.bytes()` instead. (Writing `let s = buf.reader().bytes()` doesn't work because the returned reader would not live long enough). This is admittedly only a minor annoyance.
* More importantly, I have a struct containing a RingBuf and I would like to implement a method returning a slice into that RingBuf. I can't do that, though, since the slice I want to return cannot outlive the `RingBufReader` that I have to create in that method. This is a bigger problem than the first one, since I don't think I can work around it.

I'd be interested in an explanation of the necessity of the newtypes or any thoughts you have on the subject. If you think we could get rid of them, I'd be happy to do the implementation work.

carllerche:
The answer is yes. Having separate types is a relic of an old limitation that has since been fixed. I would accept a PR that does the work :smile:

carllerche:
Fixed by #5

PR #5, opened 2015-02-24
Remove RingBufReader and RingBufWriter types

fhartwig:
As discussed in #4, this removes `RingBufReader` and `RingBufWriter`.
The only drawback of this change is that you now have to use UFCS to disambiguate the method names in certain circumstances (specifically, if you have both `Buf` and `MutBuf` in scope and call `advance` or `remaining` on a `RingBuf`).
I don't see any way to avoid this, except giving `Buf`'s and `MutBuf`'s methods distinct names, but I also don't really think this is a big problem.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -6,7 +6,7 @@
 pub use byte_buf::{ByteBuf, ROByteBuf, MutByteBuf};
 pub use byte_str::{SeqByteStr, SmallByteStr, SmallByteStrBuf};
 pub use bytes::Bytes;
-pub use ring::{RingBuf, RingBufReader, RingBufWriter};
+pub use ring::RingBuf;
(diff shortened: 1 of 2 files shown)


fhartwig:
I've also fixed the the affected places in mio, but since the pull request needs to include the new `bytes` version number, it doesn't make sense to send the pull request until the new version of bytes is on crates.io.

carllerche:
Sorry for the delay. I wanted to get MIO reform done before tackling this and I guess I forgot about it :( I will take care of it now.

carllerche:
Woops, I just noticed in your comment that you said that you also updated mio (I already pushed the changes to mio myself :cry:). I guess I am tired!

Thanks for the PR though! Merged 6f00c6117b72189018866b18bc839600021a479e

PR #9, opened 2015-03-25
Fix panics on operations on 0-capacity RingBuf

fhartwig:
These two methods currently panic when called on a `RingBuf` with `cap = 0` (not that that would be all that useful ;-)


carllerche:
Thanks!

PR #15, opened 2015-04-08
Fix Buf and MutBuf documentation

fhartwig:
Fix some documentation that was either outdated or c&p'ed incorrectly.

One thing I noticed while looking at this: `BufError` doesn't seem to be used at all. Maybe it could be removed?


carllerche:
Thanks. Merged: e22415b4162e3de34b7b7f6ae5aa79a2d764acef

PR #16, opened 2015-04-08
Fix bugs in RingBuf's io::Read and io::Write instance

fhartwig:
The implementations of `read` and `write` each use the wrong `has_remaining` method, causing writes into an empty buffer and reads from a full buffer to fail.
I'm not entirely sure why the `has_remaining` calls are there at all, since (as far as I can tell) `read_slice` and `write_slice` already handle those cases. I'm leaving them in on the assumption that I'm missing something.

### DIFF
--- a/src/buf/ring.rs
+++ b/src/buf/ring.rs
@@ -162,7 +162,7 @@ impl MutBuf for RingBuf {
 
 impl io::Read for RingBuf {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        if !MutBuf::has_remaining(self) {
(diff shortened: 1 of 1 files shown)


carllerche:
Ah yes... I will merge. It would be super awesome if you could help out w/ some tests too :smile:

carllerche:
Merged: c4546493eaad57e69e37aa4205a08000bcf82153

fhartwig:
@carllerche Yeah, absolutely. I wrote a bunch of test when debugging this, I just haven't gotten around to cleaning them up yet. Thanks for reminding me.

PR #18, opened 2015-04-10
Make buf and str modules public

fhartwig:
Commit ac427665353318bd54b61afc2de6dbb1df9f312e seems to have broken part of the documentation. Right now there is no way to access the docs of any of the structs in the `buf` and `str` modules. Making those modules public fixes this.
(It's entirely possible that there is a better way to fix the docs, but rustdocs confuses me).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,8 +3,8 @@
 #![deny(warnings)]
 
 pub mod alloc;
-mod buf;
-mod str;
+pub mod buf;
+pub mod str;
 
 pub use buf::{
     Buf,


carllerche:
This seems like a rustdoc bug. I am talking w/ @alexcrichton about it.

carllerche:
I filed an issue: https://github.com/rust-lang/rust/issues/24296

I'm not sure if we should fix in the mean time...

fhartwig:
I'm not sure either. I guess that depends on when you think this will be fixed.
I've been trying to find a workaround, but I haven't found one that works yet. I'll try to make some time this weekend to look into this issue, but I'm not very optimistic.

carllerche:
well, a large PR w/ many rustdoc fixes landed recently, so maybe let's wait to see if it handles the issue?

fhartwig:
Looks like that didn't help :-/

carllerche:
Closing this for now. Hopefully rustdoc will get a fix one  day :(

PR #19, opened 2015-04-11
add mark/reset feature to ByteBuf and RingBuf

danburkert:
This implements the mark/reset functionality we discussed on irc a few days ago.  For now `reset` will panic if `mark` has not been previously called.


carllerche:
Looks good! thanks :smile:

PR #27, opened 2015-06-03
RingBuf: Fix edge cases with mark/reset.

### LINKED ISSUE #25: RingBuf::reset sets full buffers to empty
Calling `RingBuf::reset` on a full buffer causes the buffer to become empty, due to the recalculation of `len` from `pos` and `mark`, as demonstrated by this test case:

```
let mut buf = RingBuf::new(8);
buf.write(&[1, 2, 3, 4, 5, 6, 7, 8]).unwrap();
assert_eq!(MutBuf::remaining(&buf), 0);
buf.mark();
buf.reset();
assert_eq!(MutBuf::remaining(&buf), 0);
```

### LINKED ISSUE #26: RingBuf marks aren't cleared correctly on write
When writing to a `RingBuf` with a mark set, the conditions for clearing the mark are not correct. For example, the following test case should succeed but it panics due to the mark being cleared when it shouldn't be:

```
let mut buf = RingBuf::new(8);
buf.write(&[1, 2, 3, 4, 5, 6, 7]).unwrap();
buf.mark();
buf.write(&[8]).unwrap();
buf.reset();

let mut buf2 = [0; 8];
buf.read(&mut buf2).unwrap();
assert_eq!(buf2, [1, 2, 3, 4, 5, 6, 7, 8]);
```

jamesjb:
- This fixes #25 and fixes #26.

- Move the mark state into a `Mark` enum.

- Store both `pos` and `len` inside the mark because both
  are needed to distinguish between full/empty states. This
  fixes `reset` setting `len` to zero when the buffer was
  full when marked.

- Clear the mark when the length of the marked input range
  exceeds the capacity. Fixes bugs with the mark not being
  cleared correctly on writes.

- Add test cases.

### DIFF
--- a/src/buf/ring.rs
+++ b/src/buf/ring.rs
@@ -1,6 +1,11 @@
 use {alloc, Buf, MutBuf};
 use std::{cmp, fmt, io, ptr};
 
+enum Mark {
+    NoMark,
+    At { pos: usize, len: usize },
+}
+
 /// Buf backed by a continous chunk of memory. Maintains a read cursor and a
 /// write cursor. When reads and writes reach the end of the allocated buffer,
 /// wraps around to the start.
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks!

PR #30, opened 2015-07-03
Add resume() method to ByteBuf that restores prior write position.

jamwt:
A situation we're running into is:

```rust
let mut input = ByteBuf::mut_with_capacity(8192);
read(&mut input);
let mut so_far = input.flip();
let done = parse_wire_thingie(so_far.bytes());
if !done {
    // we need more..
    input = so_far.resume();
    read(&mut input);
}
```

.. essentially, if we don't hit a whole "thingie", or find the sentinel, or whatever, we want to go back to appending more input on the end of the buffer.

This `resume()` method lets you do that; it's like `flip()`, but it puts you back to the end of the written section.

### DIFF
--- a/src/buf/byte.rs
+++ b/src/buf/byte.rs
@@ -88,6 +88,14 @@ impl ByteBuf {
         buf
     }
 
+    /// Flips the buffer back to mutable, resetting the write position
+    /// to the byte after the previous write.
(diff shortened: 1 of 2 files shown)


carllerche:
This looks good to me.

jamwt:
Awesome!

jamwt:
ty sir

PR #33, opened 2015-08-08
Allow setting the limit of a mutable buffer

thijsc:
I ran into a use case where I needed the buffer to have the exact limit of the size of a message I'm expecting to write to it. It seems like setting the limit independently from the capacity makes sense, or do you maybe consider this an implementation detail you don't want to expose?


thijsc:
Closing this one, just came across ed087be8532dba1e698e57cbb99a8e5428244985 which is a better way of achieving this use case.

carllerche:
Actually closing :)

PR #35, opened 2015-08-27
Publicize buf and str modules to get the docs generated correctly

bonifaido:


carllerche:
Sorry fro the delay. This change is on master!

PR #36, opened 2015-08-31
Add Ringbuf clear method

bonifaido:
This is for reseting the whole internal state to the initial version.

### DIFF
--- a/src/buf/ring.rs
+++ b/src/buf/ring.rs
@@ -14,6 +14,7 @@ enum Mark {
 pub struct RingBuf {
     ptr: alloc::MemRef,  // Pointer to the memory
     cap: usize,          // Capacity of the buffer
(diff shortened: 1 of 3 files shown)


carllerche:
Sorry for the delay.

It seems like there are more than one things being added with this PR.

The new `clear()` function looks good. I would pull that in.

However, I am not sure about `cap_dec`. I do see the value, but I don't think that it is worth having both `cap` & `cap_dec` stored in the struct since one can easily be computed from the other. The ability to do bit masking is definitely a pro. It would probably be worth investigating if `cap` can be fully removed in favor of `capacity - 1`

carllerche:
Closing this as #37 was merged. If anything else from this PR is relevant, please open a new one :)

PR #39, opened 2015-09-16
Fix test_vec_as_mut_buf

fhartwig:
The amount of memory allocated by `Vec::reseve` seems to have changed so that this test was broken on nightly (and will be broken on stable as of tomorrow). I've adapted the numbers in the test to fix this for now, but we should probably find a way to test this that's less fragile / doesn't rely on implementation details of `Vec`. Or we could `Vec::reserve_exact` instead of `reserve` in the `MutBuf` impl for `Vec<u8>`, if you want to have control over the amount of memory allocated.

### DIFF
--- a/test/test_buf.rs
+++ b/test/test_buf.rs
@@ -32,7 +32,7 @@ pub fn test_vec_as_mut_buf() {
     assert_eq!(buf.remaining(), usize::MAX);
 
     unsafe {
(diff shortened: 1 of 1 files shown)


carllerche:
I fixed this on master by loosening the test so that it would pass on Rust 1.1+.

Issue #24, opened 2015-05-20
Need method to mutably access already written parts of MutByteBuf


sdroege:
It would be convenient to have something like MutByteBuf::mut_bytes() that gives access to the already written part of the byte buffer instead of the remaining memory area. This would allow to modify something that was received from the network, and potentially forwarding it to some other socket, without having to copy everything.

carllerche:
Definitely an interesting thought. The use case that you have should definitely be supported. I've been mulling around some other use cases recently and will see if I can try to combine all of them and write up some new APIs to handle them... though, my initial thought is a `MutByteBuf::mut_bytes()` API is correct.

sdroege:
FWIW, code where this would be useful is here: https://github.com/sdroege/ptp-clock-reflector/blob/master/src/main.rs

Ideally I'd like to allocate the buffer only once and store it inside the struct in an Option<>, take() it from there in the beginning of readable(), receive/mutate/send, flip it back to a mutable, empty buffer and store it inside the struct again

carllerche:
I would take a `MutByteBuf::mut_bytes()` PR :)

carllerche:
Handled w/ refactor

PR #42, opened 2015-09-26
Implement Deref for Take

fhartwig:
This makes using `Take` a bit less noisy because you can call `buf.foo()` instead of `buf.get_ref().foo()`. It also make switching e.g. from `ByteBuf` to `Take<ByteBuf>` easier because you don't have to insert `get_ref()` everywhere.

### DIFF
--- a/src/buf/take.rs
+++ b/src/buf/take.rs
@@ -1,5 +1,5 @@
 use buf::{Buf, MutBuf};
-use std::{cmp, io};
+use std::{cmp, io, ops};
 
 #[derive(Debug)]
 pub struct Take<T> {
@@ -75,5 +75,13 @@ impl<T: MutBuf> MutBuf for Take<T> {
         let cnt = cmp::min(cnt, self.limit);
         self.limit -= cnt;
         self.inner.advance(cnt);
+    }
+}
+
+impl <T> ops::Deref for Take<T> {
+    type Target = T;
+
+    fn deref(&self) -> &T {
+        &self.inner
     }
 }


carllerche:
Could you explain why you often want to get access to the inner type while it is wrapped w/ `Take`? I believe that implementing `Deref` on `Take` would not be idiomatic for Rust as well as the fact that `std::io::Take` doesn't implement `Deref`.

Finally, this version of `Take` is hopefully only temporary until Rust's `std` version fixes a couple issues.

fhartwig:
I have various places where I read into a `Take<MutByteBuf>` and use `MutByteBuf::bytes()` to inspect the data. Having this `Deref` instance would mean that I don't have to keep track whether I'm dealing with a `MutByteBuf` or a `Take<MutByteBuf>`.
This isn't very important though, and I generally share your scepticism regarding `Deref`, so feel free to close this if you don't think it is a good idea. Re: std `Take` not implementing `Deref`: it also doesn't have `get_ref`, `get_mut` or `into_inner`, so it looks like you're not meant to access the inner value at all.

While we're on the topic of implementing `Deref` on things: what do you think about implementing e.g. `ByteBuf` `deref`ing to `&[u8]`. This seems fairly natural to me, and there is precedent for that in `std` (e.g. the `Deref` instance for `Vec`).

carllerche:
What's the conceptual difference between `Deref` and `AsRef`? This may work for for `ByteBuf` but probably not others.

fhartwig:
Well, `Deref` allows you to e.g. use a `&ByteBuf` wherever  a `&[u8]` is expected, while `AsRef` just gives us a way to convert a `&ByteBuf` to `&[u8]. I mostly want the former because that would allow me to use `&[u8]`'s indexing and slicing operators on a buffer directly. (Incidentally, I think implementing `DerefMut` on buffers would be a nice way to solve https://github.com/carllerche/bytes/issues/24).

I agree that this probably shouldn't be done for all buffers. In particular, I think that implementing this on `RingBuf` would lead to problems or at least be pretty non-intuitive.

carllerche:
Interesting... I'll take a PR for `Deref` on `ByteBuf`. I will definitely have to think more about `DerefMut`.

bryce-anderson:
Is there still interest in a `Deref` impl?
* `impl<T: Buf> Deref for T` would just be sugar for `Buf::bytes()`
* `impl<T: MutBuf> DerefMut for T` would be sugar for the `MutBuf::mut_bytes()` (although `MutBuf::mut_bytes()` is marked as unsafe for some reason).

#24 proposed a `MutByteBuf::mut_bytes`: does that mean you're only interested in `mut_bytes` for the concrete type `MutByteBuf`?

fhartwig:
@bryce-anderson `MutBuf::mut_bytes` is marked as unsafe because it returns a slice of (potentially) uninitialised memory. Reading from uninitialised memory is undefined behaviour in rust, so the caller has to make sure that the memory is always written to before it is read.

bryce-anderson:
@fhartwig, that makes sense. I'll put a PR together for `MutByteBuf` (eg #24) and add the `Deref` and `DerefMut` impls. Its probably best to split that into 2 PRs.

Sorry for hijacking this PR discussion.

carllerche:
Better late than never?

Thanks for the original PR :) The work / feedback is much appreciated.

After thinking about adding `Deref` to `Take`, I'm going to close this without merging. The danger of adding `Deref` is that implicitly calling the internal fns is not an expected behavior because `Take` is supposed to enhance the behavior of the inner value. Implementing `Deref` may cause the enhanced behavior to be bypassed unexpectedly. An explicit call to `get_ref` mitigates this unexpected behavior.

Thanks again for the PR :)

PR #43, opened 2015-09-27
Add constructor to MutByteBuf

fhartwig:
Strictly speaking, this is redundant because you could always use `ByteBuf::mut_with_capacity`, but I still think this is worth adding because it's possible to use `MutByteBuf` without using `ByteBuf` at all. Having to go through `ByteBuf` to construct a `MutByteBuf` is a bit weird in these cases (and also not obvious from looking at the `MutByteBuf` docs).

### DIFF
--- a/src/buf/byte.rs
+++ b/src/buf/byte.rs
@@ -23,14 +23,9 @@ pub struct ByteBuf {
 impl ByteBuf {
     /// Create a new `ByteBuf` by copying the contents of the given slice.
     pub fn from_slice(bytes: &[u8]) -> ByteBuf {
-        let mut buf = ByteBuf::mut_with_capacity(bytes.len());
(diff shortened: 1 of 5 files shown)


carllerche:
I would be ok but I think that `with_capacity`. It may be worth getting rid of `ByteBuf::mut_with_capacity`.

fhartwig:
Ok, I'll change the name an remove `mut_with_capacity`. I mostly went with `new` for consistency with the constructor for `RingBuf`.

carllerche:
`RingBuf` probably should be changed too... I am also trying to match Rust's `std`.

fhartwig:
Ok, I've amended my commit.

carllerche:
Thanks! I added this here 38abb8074b87f226c32213ebb1bee7752b43278c

Issue #46, opened 2016-08-26
Buffer overflow


stbuehler:
Buffer overflow in `impl<'a> Sink for &'a mut Vec<u8>`

```rust
	let mut sink: Vec<u8> = Vec::new();
	sink.reserve(16);
	println!("Capacity: {}", sink.capacity());
	let mut source = Cursor::new(b"0123456789abcdef0123456789abcdef");
	sink.copy_from(&mut source);
	println!("Length: {}, Capacity: {}", sink.len(), sink.capacity());
```

prints:
```
Capacity: 16
Length: 32, Capacity: 16
```

The error is in the call to reserve: you don't reserve "more capacity", instead you are only guaranteed there is enough space to insert the requested number of elements, i.e. you need to base it on length instead of capacity.

Is there a really good reason why you need `unsafe` here?

```rust
	self.resize(buf.remaining(), 0);
	buf.read_slice(&mut self[..]);
```

Is not zeroing the buffer really that performance critical to risk buffer overflows like this one?

PR #47, opened 2016-09-02
Fix buffer overflow in impl<'a> Sink for &'a mut Vec<u8> (#46)

stbuehler:


carllerche:
Thanks, merged: f693e038d9ad3042c0a68bb3ef609117b1d7a86a

PR #48, opened 2016-09-13
fix dropping front block in BlockBuf

fafhrd91:
if remaining bytes is 0 for front block, then BlockBuf::drop always pops front block.


carllerche:
Ah, good call. Thanks.

PR #50, opened 2016-09-20
added BufferPool trait

rrichardson:


carllerche:
Closing due to inactivity. I like the idea of a buffer pool trait, but I think it should be experimented with out of the bytes crate and we can consider moving it in once the design space has been explored.

Thanks for the PR though!

PR #52, opened 2016-10-28
Remove println!() from RingBuf implementation

hjr3:


carllerche:
Thanks for the PR.

`RingBuf` is going to be moved out of `bytes` crate into `bytes-more` for the time being to let it incubate. #53 will be removing it in order to focus on a smaller API set in hopes to be able to reach stability sooner.

I included this diff when moving `RingBuf` out to here: https://github.com/carllerche/bytes-more/blob/master/src/ring.rs

PR #58, opened 2017-02-15
Provide two versions of drain_to and split_off

carllerche:
* `drain_to` and `split_off` take &self and return Bytes.
* `drain_to_mut` and `split_off_mut` take &mut self and return BytesMut


carllerche:
This was a bad idea.

PR #60, opened 2017-02-28
Add vectored support to Buf and BufMut

carllerche:
cc @alexcrichton

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ categories = ["network-programming", "data-structures"]
 
 [dependencies]
 byteorder = "1.0.0"
+iovec = { git = "https://github.com/carllerche/iovec" }
 
 [dev-dependencies]
 tokio-core = "0.1.0"
--- a/src/buf/buf.rs
+++ b/src/buf/buf.rs
@@ -1,5 +1,6 @@
 use super::{Take, Reader, Iter, FromBuf};
 use byteorder::ByteOrder;
+use iovec::IoVec;
 
 use std::{cmp, io, ptr};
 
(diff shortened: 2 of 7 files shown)


alexcrichton:
A few thoughts:

* I wonder if these methods should be the "required to implement" methods? They're certainly wordier than the `bytes` method but do we envision all that many impls of these traits?
* Maybe a subslice could be returned instead of a `usize`? Something like:

  ```rust
  fn foo<'a, 'b>(&'a self, slice: &'b mut [&'a IoVec]) -> &'b mut [&'a IoVec];
  ```

  That's a lot of lifetimes but makes the caller's life a little easier I think.

* I wonder if we should support offsets of some form? I'm thinking that you don't really have any way of knowing how large of a slice to pass in, but maybe that's not so bad. Do you think this'll end up being a problem?

carllerche:
> I wonder if these methods should be the "required to implement" methods? They're certainly wordier than the bytes method but do we envision all that many impls of these traits?

I don't think it matters a ton. 0.4 is released. The way it is means that you can "quickly" impl Buf / BufMut w/o depending on `IoVec` and `IoVec` is kept out of the bytes documentation (no re-export). There won't be a ton of impls so it doesn't matter too much.

> Maybe a subslice could be returned instead of a `usize`?

I don't have a strong opinion either way.

> I wonder if we should support offsets of some form? I'm thinking that you don't really have any way of knowing how large of a slice to pass in, but maybe that's not so bad. Do you think this'll end up being a problem?

I'm not hugely following... my interpretation is that there is no way to know how many distinct chunks are remaining in the `Buf`. I don't think that the `Buf` itself could always know w/o extra work. I'm not hugely concerned about this as you can always supply a large number of stack allocated IoVec refs.

carllerche:
A quick note re: slice return value: Here is an implementation of chaining bufs (similar to iterator chaining):

```rust
struct Chain<T, U> {
    a: T,
    b: U,
}

fn bytes_vec<'a>(&'a self, dst: &mut [&'a IoVec]) -> usize {
    if self.a.remaining() {
        let mut n = self.a.bytes_vec(dst);

        if n < dst.len() {
            n += self.b.bytes_vec(&mut dst[n..]);
        }

        n
    } else {
        self.b.bytes_vec(dst)
    }
}
```

It still can be implemented w/ the slice ret val, but I kind of like this way :)

Either way...

alexcrichton:
Discussed on IRC, I'm 👍 with this as-is. I'll open follow-up issues for 0.5

PR #68, opened 2017-03-01
Chain

carllerche:


src/buf/chain.rs
```diff
@@ -0,0 +1,241 @@
...
+
+            cnt -= a_rem;
+        }
+
+        self.b.advance(cnt);
+    }
+
+    fn bytes_vec<'a>(&'a self, dst: &mut [&'a IoVec]) -> usize {
+        if self.a.has_remaining() {
+            let mut n = self.a.bytes_vec(dst);
```
alexcrichton: Maybe this could unconditionally be called to avoid the extra branch?

PR #71, opened 2017-03-02
Improved reserve

carllerche:
`BytesMut::reserve` should avoid small allocations

This change tracks the original capacity requested when `BytesMut` is first created. This capacity is used when a `reserve` needs to allocate due to the current view being too small. The newly allocated buffer will be sized the same as the original allocation.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -365,12 +365,7 @@ impl Bytes {
     pub fn new() -> Bytes {
         Bytes {
             inner: Inner2 {
-                inner: Inner {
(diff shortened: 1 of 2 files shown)


alexcrichton:
Heh I think you may be on your own with this one. I don't mind trying to review here, but it'll take me a lot longer to learn the lib in detail and such. I'd be fine if you'd prefer to just merge in the meantime.

carllerche:
I'll merge. This shouldn't change the public API. Reserve is still
documented as being able to allocate more than requested.

Doing a full review at some point would be good.

PR #82, opened 2017-03-19
Reduce is_inline calls in Bytes's BufMut impl

### LINKED ISSUE #81: BytesMut's BufMut impl is much slower than Vec's
Currently, it is ~2x slower. A cursory investigation reveals that there are way too many `is_inline` checks. Removing them brings the impls back on par.

carllerche:
Fixes #81

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1117,18 +1117,12 @@ impl BufMut for BytesMut {
 
     #[inline]
     unsafe fn advance_mut(&mut self, cnt: usize) {
-        let new_len = self.len() + cnt;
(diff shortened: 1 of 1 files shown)


carllerche:
We'll probably not want to merge this as is, and instead to a more systematic review of the BytesMut fns.

arthurprs:
Maybe this should be merged?

carllerche:
Well, this isn't super critical and I was hoping to do a more systematic pass / optimization.

seanmonstar:
This is no longer relevant.

PR #93, opened 2017-03-24
Flag `Deref` methods as `#[inline]`

alexcrichton:


alexcrichton:
Note that as `Bytes` and `BytesMut` are concrete types these methods are never otherwise inlined so I figured that this could help in those situations (and I saw them in profiles where this PR caused those to be erased from the profile)

PR #96, opened 2017-03-28
serde support behind serde feature

arthurprs:
I needed it in a project so I wrote a straightforward implementation and added it behind the serde_impl feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ categories = ["network-programming", "data-structures"]
 [dependencies]
 byteorder = "1.0.0"
 iovec = "0.1"
+serde = { version = "1.0", optional = true }
 
 [dev-dependencies]
-tokio-core = "0.1.0"
+serde_test = "1.0"
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -1,0 +1,82 @@
+extern crate serde;
+
+use std::{cmp, fmt};
+use self::serde::{Serialize, Serializer, Deserialize, Deserializer, de};
+use super::{Bytes, BytesMut};
+
+macro_rules! serde_impl {
+    ($ty:ident, $visitor_ty:ident) => (
+        impl Serialize for $ty {
+            #[inline]
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+                where S: Serializer
+            {
+                serializer.serialize_bytes(&self)
+            }
+        }
+
+        struct $visitor_ty;
+
+        impl<'de> de::Visitor<'de> for $visitor_ty {
+            type Value = $ty;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("byte array")
+            }
+
+            #[inline]
+            fn visit_seq<V>(self, mut seq: V) -> Result<Self::Value, V::Error>
+                where V: de::SeqAccess<'de>
+            {
+                let len = cmp::min(seq.size_hint().unwrap_or(0), 4096);
+                let mut values = Vec::with_capacity(len);
+
+                while let Some(value) = try!(seq.next_element()) {
+                    values.push(value);
+                }
+
+                Ok(values.into())
+            }
+
+            #[inline]
+            fn visit_bytes<E>(self, v: &[u8]) -> Result<Self::Value, E>
+                where E: de::Error
+            {
+                Ok($ty::from(v))
+            }
+
+            #[inline]
+            fn visit_byte_buf<E>(self, v: Vec<u8>) -> Result<Self::Value, E>
+                where E: de::Error
+            {
+                Ok($ty::from(v))
+            }
+
+            #[inline]
+            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+                where E: de::Error
+            {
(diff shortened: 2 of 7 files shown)


carllerche:
cc @alexcrichton

alexcrichton:
I initially thought this was a good idea but it's an important point that we'd be hitching the version of the entire futures stack to the serde bandwagon. I'm not sure if we're ready to make that committment yet?

arthurprs:
Will that ever go away though?

alexcrichton:
Sure yeah, once serde hits 1.0

arthurprs:
Updated for serde 1.0 :tada:

Cargo.toml
```diff
@@ -19,9 +19,11 @@ exclude       = [
...
+serde_impl = ["serde", "serde_test"]
+
 [dependencies]
 byteorder = "1.0.0"
 iovec = "0.1"
-
-[dev-dependencies]
-tokio-core = "0.1.0"
+serde = { version = "1.0", optional = true }
+serde_test = { version = "1.0", optional = true }
```
dtolnay: `serde_test` should be a dev-dependency. That way you can call the serde feature simply "serde" instead of "serde_impl".

src/serde.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        impl<'de> de::Visitor<'de> for $visitor_ty<$ty> {
+            type Value = $ty;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("byte array")
+            }
+
+            #[inline]
+            fn visit_unit<E>(self) -> Result<Self::Value, E>
```
dtolnay: Please remove this method. Questionable implicit conversions like this were removed in Serde 1.0.

src/serde.rs
```diff
@@ -0,0 +1,94 @@
...
+        impl Serialize for $ty {
+            #[inline]
+            fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+                where S: Serializer
+            {
+                serializer.serialize_bytes(&self)
+            }
+        }
+
+        struct $visitor_ty<$ty> {
```
dtolnay: The type parameter and PhantomData should not be necessary here. This can be `struct $visitor_ty;`.

arthurprs:
@dtolnay thanks for the review!

arthurprs:
Travis is failing because the dev-dependency on serde-test. I can either revert the feature to be "serde_impl" and have both serde and serde_test crates behind it. OR we can bump the minimum version required for bytes.

dtolnay:
I don't think that is why it is failing.

```
   Compiling serde v1.0.1
serde-1.0.1/src/ser/impls.rs:26:44: 26:49 error: expected one of `!`, `,`, `.`, `::`, `?`, `{`, or an operator, found `as`
serde-1.0.1/src/ser/impls.rs:26                 serializer.$method(*self $($cast)*)
                                                                           ^~~~~
serde-1.0.1/src/ser/impls.rs:26:44: 26:49 error: expected one of `!`, `.`, `::`, `;`, `?`, `{`, `}`, or an operator, found `as`
serde-1.0.1/src/ser/impls.rs:26                 serializer.$method(*self $($cast)*)
                                                                           ^~~~~
serde-1.0.1/src/ser/impls.rs:26:44: 26:49 error: expected one of `!`, `,`, `.`, `::`, `?`, `{`, or an operator, found `as`
serde-1.0.1/src/ser/impls.rs:26                 serializer.$method(*self $($cast)*)
                                                                           ^~~~~
serde-1.0.1/src/ser/impls.rs:26:44: 26:49 error: expected one of `!`, `.`, `::`, `;`, `?`, `{`, `}`, or an operator, found `as`
serde-1.0.1/src/ser/impls.rs:26                 serializer.$method(*self $($cast)*)
                                                                           ^~~~~
serde-1.0.1/src/ser/impls.rs:140:39: 140:43 error: expected expression, found `01`
serde-1.0.1/src/ser/impls.rs:140             impl<T> Serialize for [T; $len]
                                                                                                                       ^~~~
error: Could not compile `serde`.
```

I would accept a PR to fix this in Serde to support 1.10.0, or we say that `bytes` works on ancient Rust but Serde support requires a less ancient Rust. In that case you need to figure out how to disable the `--feature serde` on the oldest Travis builds.

arthurprs:
Maybe I didn't phrase that correctly. It fails because serde_test (and thus serde) is always included in test builds and the later doesn't work on 1.10. As far as I know there's no way of making dev-dependencies optional.

dtolnay:
Oh I see, yes you're right. Maybe we only `cargo build` on 1.10.0 and don't run tests?

carllerche:
The min version can be bumped to 1.14 if that helps.

carllerche:
I guess 1.17 is being released tomorrow, so it could be bumped to 1.15.

carllerche:
@arthurprs Sorry, would you mind making the version bump as a separate PR? I would like to keep the bump in the git history. Thanks

arthurprs:
Sure thing, let's just see what travis thinks about it.

arthurprs:
Re-triggering Travis.

carllerche:
Thanks

PR #101, opened 2017-04-05
implement put_u8 for BytesMut

seanmonstar:
Benching before and after shows that this implementation is about twice as fast as going through `put_slice`.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1164,6 +1164,16 @@ impl BufMut for BytesMut {
             self.advance_mut(len);
         }
     }
+
+    #[inline]
+    fn put_u8(&mut self, n: u8) {
+        self.inner.put_u8(n);
+    }
+
+    #[inline]
+    fn put_i8(&mut self, n: i8) {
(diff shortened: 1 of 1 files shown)


carllerche:
You could probably make it even faster by only having one `is_inline()` check :) You can roughly see what I mean here: https://github.com/carllerche/bytes/pull/82

Do you want to try that out?

seanmonstar:
Yes, reducing the `is_inline` to one call makes it a little over 2x faster than this proposed implementation, for a total ~4x faster...

src/bytes.rs
```diff
@@ -1463,6 +1475,21 @@ impl Inner {
         }
     }
 
+    /// Insert a byte into the next slot and advance the len by 1.
+    #[inline]
+    unsafe fn put_u8(&mut self, n: u8) {
```
carllerche: Why is this function unsafe? While the implementation requires `unsafe`, I don't think that unsafety leaks out?

seanmonstar:
Laziness :). Fixed.

PR #121, opened 2017-05-18
Introduce more bridges between Bytes and BytesMut 🌉

nox:


carllerche:
`into_owned` functionality is already provided by `impl From<Bytes> for BytesMut`.

Regarding the `get_mut`, could you describe a bit more what the use case is?

Adding the borrow impl seems good. Could you move that to its own PR?

nox:
I will make the separate PR for `Borrow`.

As for the use case for `get_mut`, it's nonobvious because I somehow confused myself and made `into_owned` instead of `make_mut`, which was the method I wanted to make initially. The method `make_mut` would call `get_mut`, which I made public just because it is on `Arc<T>` too.

carllerche:
Do you agree that `into_owned` is the same as the `From` implementation? If so, maybe remove it from the PR and then we can focus on `get_mut`.

carllerche:
Closing due to inactivity.

nox:
@carllerche I remember the use case for get_mut: there is no way to get own a `&mut Bytes` and get a `&mut BytesMut` from that.

nox:
I'll reopen a new PR for `get_mut`, but IMO `into_owned` should be provided, if only for consistency with the `Cow<str>` type.

carllerche:
I like the idea of mirroring APIs in `std`. However, I'm not 100% sure about copying the `into_owned` terminology.

`Cow` is an enum of `Borrowed` / `Owned`. As such, `into_owned` makes sense as it converts the value to the owned variant.

In `bytes`, it's either `Bytes` or `BytesMut`, and there already is `try_mut`. So, adding `into_owned` would clash in terms of naming.

Maybe `into_mut` would work?

nox:
I think `into_mut` sounds slightly to close to `to_mut` now, hah. Your point about `Cow` being an enum makes sense, so I am with either choice you make about introducing `into_owned` or not, given that the `From` impl exists. Maybe we should just document the various ways to go from `Bytes` to `BytesMut` more prominently?

Issue #126, opened 2017-05-23
Implement ExactSizeIterator for Iter<T: Buf>


brianwp:
An std::iter::[ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) is "an iterator that knows its exact length". 
Implementing ExactSizeIterator for Iter<T: Buf> is trivial because it already returns the exact size in its implementation of size_hint()--it just needs an empty trait impl.

carllerche:
Fixed by #127

PR #127, opened 2017-05-23
impl ExactSizeIterator for Iter<T: Buf>

brianwp:
Pull request for issue #126.
I added some unit tests, but they may be unnecessary because the method implementations are trivial.


carllerche:
Thanks!

PR #133, opened 2017-06-07
Faster From<[u8]> for BytesMut, remove panic in fmt::Write

seanmonstar:
Benchmark change for `From<[u8]>`:

Before:
```
test from_long_slice      ... bench:       45 ns/iter (+/- 2) = 2844 MB/s
```
After:
```
test from_long_slice      ... bench:       40 ns/iter (+/- 2) = 3200 MB/s
```

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1484,11 +1484,17 @@ impl Borrow<[u8]> for BytesMut {
 }
 
 impl fmt::Write for BytesMut {
+    #[inline]
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        BufMut::put(self, s);
-        Ok(())
+        if self.remaining_mut() >= s.len() {
+            self.put_slice(s.as_bytes());
+            Ok(())
+        } else {
+            Err(fmt::Error)
+        }
     }
 
+    #[inline]
     fn write_fmt(&mut self, args: fmt::Arguments) -> fmt::Result {
         fmt::write(self, args)
     }
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1418,10 +1418,7 @@ impl<'a> From<&'a [u8]> for BytesMut {
                 }
             }
         } else {
-            let mut buf = BytesMut::with_capacity(src.len());
-            let src: &[u8] = src.as_ref();
-            buf.put(src);
-            buf
+            BytesMut::from(src.to_vec())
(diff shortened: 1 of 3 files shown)


src/bytes.rs
```diff
@@ -1488,7 +1485,7 @@ impl Borrow<[u8]> for BytesMut {
 
 impl fmt::Write for BytesMut {
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        BufMut::put(self, s);
+        self.extend_from_slice(s.as_bytes());
```
carllerche: This has different behavior (`extend_from_slice` grows the buffer but `put` panics on full).

That said, this is probably fine? Thoughts?

seanmonstar: Hm, that it doesn't grow when being written to is a little surprising, no? Most other containers that can be written to do so...

carllerche: My opinion has been that you don't want your buffer to implicitly grow. This is due to `BufMut` implementations having an explicit capacity.

It could be that this should be changed... but there will be implications w/ the `BufMut` impl that would need to be thought through.

seanmonstar: If a `BytesMut` should be treated like a `&mut [u8]`, then I propose that the `fmt::Write` work the same way.

While there isn't a `fmt::Write` impl for `&mut [u8]`, there is an `io::Write`, (so you can do `write!(slice_mut, "{}", foo)`. The `Write` impl writes all it can to the end of the buffer, and then additional writes return `Ok(0)`. The `io::Write::write_fmt` default method tries to use `write_all`, so it will eventually return a `WriteZero` error if the mutable slice doesn't have any more room. In turn, this is translated to `fmt::Error`. This means that it can error, but it **doesn't panic** unless you choose to, by unwrapping the error.

Proposed implementation:

```rust
impl fmt::Write for BytesMut {
    fn write_str(&mut self, s: &str) -> fmt::Result {
        if s.len() <= self.remaining_mut() {
            // hoping that the compiler can elide the assert in `put_slice`
            // since it's the same check
            self.put_slice(s.as_bytes());
            Ok(())
        } else {
            Err(fmt::Error)
        }
    }
}
```

seanmonstar:
The behavior is now the same as `&mut [u8]`, where if the string won't fit, a `fmt::Error` is returned.

Issue #75, opened 2017-03-16
`io::Cursor` is a bit out of place


alexcrichton:
I think this may not wish to use `io::Cursor` pervasively. This use case doesn't have much to do with `std::io`. The std version also [stores a ](https://github.com/rust-lang/rust/blob/0aeb9c12979e6da753701a798d04105b6b1a8c28/src/libstd/io/cursor.rs#L86) `u64` which isn't needed for `bytes`'s use case I believe.

alexcrichton:
I also just twitch a bit whenever I see `io::Cursor`, I sort of regret adding it to std, it doesn't seem to really pull its weight.

carllerche:
Makes sense. I mostly opted for `Cursor` to not add another type.

Another option would be to bring back `ByteBuf` or something like that?

alexcrichton:
I think a local "cursor type" (w/e the name) may be right here, but I wouldn't add it just yet (I don't know what it should look like)

est31:
> I also just twitch a bit whenever I see io::Cursor, I sort of regret adding it to std, it doesn't seem to really pull its weight.

A bit off topic, but note that I love it being in std, you can easily use it to Read over a buffer. Very interesting when you only offer a `io::Read` based API and people want to use it with their memory buffered data (best use case is tests).

carllerche:
Using `Cursor` here is definitely a mistake on my part...

`std` implements `Read` and `Write` for `&[u8]` and `&mut [u8]` respectively. We probably want to follow that pattern and implement `Buf` / `BufMut` for `&[u8]` and `&mut [u8]`.

tarcieri:
+1 on this. `io::Cursor` complicates `no_std` usage. I had to vendor it in to my PR to add `no_std` support: #135

Perhaps I should change that PR to just stop using `io::Cursor` entirely?

tarcieri:
I opened a PR which vendors `std::io::Cursor` as `buf::Cursor` here (while also reducing the position counter size to `u32`):

https://github.com/carllerche/bytes/pull/146

I'm not sure that's entirely ideal but I don't see any other path forward here. I need something like it to make progress on #135

seanmonstar:
So to clarify what will happen in v0.5:

- `impl Buf for Bytes` directly (remove `IntoBuf` of `Cursor<Self>`)
- `impl Buf for BytesMut` directly (remove `IntoBuf` of `Cursor<Self>`)
[...]

PR #135, opened 2017-06-16
no_std support (using enabled-by-default "std" feature)

tonychain:
- Changes all references to libcore features from `::std` to `::core`
- Feature gates anything dependent on `std` on the "std" feature

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,12 +17,19 @@ exclude       = [
     "bench/**/*",
     "test/**/*"
 ]
-categories = ["network-programming", "data-structures"]
+categories = ["network-programming", "data-structures", "no-std"]
 
 [dependencies]
 byteorder = "1.0.0"
-iovec = "0.1"
 serde = { version = "1.0", optional = true }
+
+[dependencies.iovec]
+version = "0.1"
+optional = true
 
 [dev-dependencies]
 serde_test = "1.0"
+
+[features]
+default = ["std"]
+std = ["iovec"]
--- a/src/buf/into_buf.rs
+++ b/src/buf/into_buf.rs
@@ -1,5 +1,6 @@
 use super::{Buf};
 
+#[cfg(feature = "std")]
 use std::io;
 
 /// Conversion into a `Buf`
@@ -55,6 +56,7 @@ impl<T: Buf> IntoBuf for T {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a [u8] {
     type Buf = io::Cursor<&'a [u8]>;
 
@@ -63,6 +65,7 @@ impl<'a> IntoBuf for &'a [u8] {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a str {
     type Buf = io::Cursor<&'a [u8]>;
 
@@ -71,6 +74,7 @@ impl<'a> IntoBuf for &'a str {
     }
 }
 
+#[cfg(feature = "std")]
 impl IntoBuf for Vec<u8> {
     type Buf = io::Cursor<Vec<u8>>;
 
@@ -79,6 +83,7 @@ impl IntoBuf for Vec<u8> {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a Vec<u8> {
     type Buf = io::Cursor<&'a [u8]>;
 
@@ -89,6 +94,7 @@ impl<'a> IntoBuf for &'a Vec<u8> {
 
 // Kind of annoying... but this impl is required to allow passing `&'static
 // [u8]` where for<'a> &'a T: IntoBuf is required.
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a &'static [u8] {
     type Buf = io::Cursor<&'static [u8]>;
 
@@ -97,6 +103,7 @@ impl<'a> IntoBuf for &'a &'static [u8] {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a &'static str {
     type Buf = io::Cursor<&'static [u8]>;
 
@@ -105,6 +112,7 @@ impl<'a> IntoBuf for &'a &'static str {
     }
 }
 
+#[cfg(feature = "std")]
 impl IntoBuf for String {
     type Buf = io::Cursor<Vec<u8>>;
 
@@ -113,6 +121,7 @@ impl IntoBuf for String {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a String {
     type Buf = io::Cursor<&'a [u8]>;
 
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -2,11 +2,13 @@ use {IntoBuf, Buf, BufMut};
 use buf::Iter;
 use debug;
 
-use std::{cmp, fmt, mem, hash, ops, slice, ptr, usize};
-use std::borrow::Borrow;
-use std::io::Cursor;
-use std::sync::atomic::{self, AtomicUsize, AtomicPtr};
-use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
+use core::{cmp, fmt, mem, hash, ops, slice, ptr, usize};
+use core::borrow::Borrow;
+use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
+use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
+
+#[cfg(feature = "std")]
+use std::io;
 
 /// A reference counted contiguous slice of memory.
 ///
@@ -320,10 +322,10 @@ struct Inner2 {
 }
 
 // Thread-safe reference-counted container for the shared storage. This mostly
-// the same as `std::sync::Arc` but without the weak counter. The ref counting
+// the same as `core::sync::Arc` but without the weak counter. The ref counting
 // fns are based on the ones found in `std`.
 //
-// The main reason to use `Shared` instead of `std::sync::Arc` is that it ends
+// The main reason to use `Shared` instead of `core::sync::Arc` is that it ends
 // up making the overall code simpler and easier to reason about. This is due to
 // some of the logic around setting `Inner::arc` and other ways the `arc` field
 // is used. Using `Arc` ended up requiring a number of funky transmutes and
@@ -767,18 +769,18 @@ impl Bytes {
 }
 
 impl IntoBuf for Bytes {
-    type Buf = Cursor<Self>;
+    type Buf = io::Cursor<Self>;
 
     fn into_buf(self) -> Self::Buf {
-        Cursor::new(self)
+        io::Cursor::new(self)
     }
 }
 
 impl<'a> IntoBuf for &'a Bytes {
-    type Buf = Cursor<Self>;
+    type Buf = io::Cursor<Self>;
 
     fn into_buf(self) -> Self::Buf {
-        Cursor::new(self)
+        io::Cursor::new(self)
     }
 }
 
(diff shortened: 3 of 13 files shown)


tarcieri:
Sorry to bring up a bikeshedding topic, but I've been looking at `use_std` vs `std` as a feature name.

@alexcrichton I chose `use_std` to match futures-rs, however per some various bikeshedding issues, I think people might be aligning around `std`.

See:

https://github.com/brson/rust-api-guidelines/issues/23
https://github.com/rust-lang-nursery/log/issues/176

alexcrichton:
Long ago I remember we discussed creating a convention for the "enable std support" feature for the `log` crate and we settled on `use_std`. I have no opinion on this so long as we are consistent. Unfotunately, nothing is consistent right now.

Cargo.toml
```diff
@@ -21,8 +21,16 @@ categories = ["network-programming", "data-structures"]
 
 [dependencies]
 byteorder = "1.0.0"
-iovec = "0.1"
 serde = { version = "1.0", optional = true }
 
+[dependencies.iovec]
```
carllerche: Ok, so this means that the vectored IO fns on the traits aren't available, which doesn't seem great (long term).

That said, they could be brought back incrementally w/o a breaking change.

Cargo.toml
```diff
@@ -21,8 +21,16 @@ categories = ["network-programming", "data-structures"]
...
+version = "0.1"
+optional = true
+
 [dev-dependencies]
 serde_test = "1.0"
+
+[features]
+default = ["use_std"]
+use_alloc = []
+use_std = ["iovec"]
```
carllerche: It seems that `use_std` implies `use_alloc`. I'm not sure if there is a way to define this in Cargo.toml.

src/buf/into_buf.rs
```diff
@@ -1,6 +1,14 @@
 use super::{Buf};
 
-use std::io;
+#[cfg(feature = "use_alloc")]
+use alloc::string::String;
+#[cfg(feature = "use_alloc")]
+use alloc::vec::Vec;
+
+#[cfg(not(feature = "use_std"))]
+use buf::Cursor;
```
carllerche: Unfortunately, given that features must be additive, I don't think this conditionally vendoring cursor is something that can be done.

This, of course, means that `IntoBuf` can't really be implemented without `std` :(

tonychain: A couple options here:

1) Gate this (and anything that depends on it) on `use_alloc`
2) Stop using `std::io::Cursor` entirely (i.e. fix #75)

Cargo.toml
tonychain: Err, other way around: `use_alloc` and `use_std` are orthogonal. `use_alloc` is intended for `no_std` use, i.e. `--no-default-features --features=use_alloc`

bytes builds without it per this PR, but with so many things disabled it isn't particularly useful

src/buf/into_buf.rs
carllerche: @alexcrichton if `buf::Cursor` is defined as a re-export of `std::io::Cursor` when `use_std` is set, is that ok for "additive" features?

alexcrichton:
Ah yeah I agree with @carllerche in that the handling of `Cursor` here, while unfortunate, I don't think is a valid usage of features. Features *must* be composable in that they purely add APIs rather than change existing ones.

I'm personally not a huge fan of using unstable crates like `alloc` because that also means that this library sometimes requires nightly and other times not. That may just be my own personal preference though.

tarcieri:
The cleanest solution re: `Cursor` is probably to just wait for #75 to get fixed (which I think targets bytes 0.5), which I think probably wouldn't be too far off from what I've already done, but using that 100% of the time as opposed to relying on `std::io::Cursor` at all.

Regarding `alloc` and nightly: to my knowledge nightly is required for any sort of `no_std` usage (since `lang_items` is needed to use `no_std`, AFAICT). The crate still builds on its own with `--no-default-features`, but `alloc` (or `std`) is really necessary to get a lot of the useful functionality.

tarcieri:
All right, took a stab at #75 here: https://github.com/carllerche/bytes/pull/146

That should get past having two different implementations of `buf::Cursor` depending on whether or not we have `use_std` enabled.

alexcrichton:
Yeah I think with `Cursor` there's not much recourse but waiting for the next major version (unfortunately). For `alloc` you're right, but the current state of affairs is that allocation without the standard library is unstable right now. This may change in the near future with the pending stabilization of allocators, but I'd personally prefer that a mixture of features for a library doesn't require it oscillate between requiring nightly or not.

tarcieri:
All `no_std` users are presently stuck on nightly. It'd be great to get the many things keeping `no_std` users on nightly stabilized, but in the meantime I'm willing to tolerate a little instability, and have some pressing needs for crates that depend on `bytes`.

Personally I would much rather deal with some changes to nightly-only features of `bytes` targeted at `no_std` users than be locked out of the ecosystem of crates that depend on `bytes`, with my only recourse to maintain a fork of `bytes` and forks of the crates that depend on `bytes`.

alexcrichton:
Oh sure that makes sense for `no_std` users but this is otherwise a stable crate, and I'd consider it a little too surprising for this crate to sometimes require nightly and sometimes not, depending on activate features.

tarcieri:
Who would be trying to swap `use_std` for `use_alloc` except for nightly-only `no_std` users?

alexcrichton:
There's use cases for no_std on stable, nightly is not the *only* consumer.

tarcieri:
I think pretty much all practical `no_std` uses need `lang_items` at the very least

alexcrichton:
Ok, well, my opinion is that this crate should not have a feature where it only uses the `alloc` crate. That's an unstable requirement in an otherwise stable crate, and can cause surprises when Cargo unions features together. My recommendation is that, if we really want it, the no_std feature of this crate should just provide the traits. I'll leave it up to @carllerche to merge if he'd like.

tarcieri:
@alexcrichton would it help if the feature were more clearly labeled to indicate its experimental/nightly nature, e.g. `nightly_alloc` or `experimental_alloc`?

alexcrichton:
I would still personally prefer to not add such a feature to the crate, but I'd again defer to @carllerche's thoughts.

dtolnay:
I know this is mentioned in the threads that were linked above but to reiterate: I strongly prefer `std` over `use_std` and I plan to push that as the one we standardize on.

If some crate has an optional dependency on `bytes`:

```toml
[package]
name = "tarcieri"

[dependencies]
bytes = { version = "0.4", optional = true }
```

then I depend on it like this:

```toml
[package]
name = "dtolnay"

[dependencies]
tarcieri = { version = "0.1", features = ["bytes"] }
```

Why should it be any different for an optional dependency on the `std` or `alloc` crates? If a crate has an optional dependency on `std`, I should be able to enable that optional dependency with `features = ["std"]` and same for `alloc`.

tarcieri:
@alexcrichton perhaps it'd be better if I started with a less ambitious PR which gates all functionality which requires allocators on the `std` feature (which, per @dtolnay, should probably be called `std`)

We can then look at what functionality is missing from `bytes` that's actually needed in the downstream crates which are consuming it (`IntoBuf` was the main one of interest for me), and whether we can handle the allocations in the downstream crates rather than in `bytes` itself.

If it turns out that's a bad approach, we can circle back on handling `no_std` allocations via liballoc in a separate PR.

Does that seem like a better way forward to you?

alexcrichton:
@tarcieri seems reasonable to me!

tarcieri:
Okay, I've updated the PR to massively reduce the scope and feature set for `no_std` usage. This should be dramatically simpler and we can circle back on allocation.

Per https://github.com/brson/rust-api-guidelines/commit/a15e953eace73b10759824a6fc7e703afd724f8c this uses "std" as the feature name (cc @dtolnay)

tonychain:
I moved all of the allocator-dependent gating to a separate PR:

https://github.com/carllerche/bytes/pull/153

carllerche:
Sorry for the delay.

At a high level, this looks good to me. This simply feature flags everything that requires std. I think @alexcrichton has some thoughts on how to organize it though to reduce the actual number of cfg lines needed...

tarcieri:
Definitely interested in ideas regarding how to simplify this sort of gating, as I've done in in over a dozen crates at this point, and would agree it doesn't feel particularly good doing this way.

src/lib.rs
```diff
@@ -70,8 +70,13 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.4")]
+#![cfg_attr(not(feature = "std"), no_std)]
```
alexcrichton: I think one of the easiest things to do to ease development of no_std libraries is to unconditionally use the `#![no_std]` attribute.

Only if the `std` feature is enabled should this have `extern crate std`, and then in that case the prelude can be imported to specific modules if necessary.

src/bytes.rs
```diff
@@ -2,11 +2,13 @@ use {IntoBuf, Buf, BufMut};
...
-use std::borrow::Borrow;
-use std::io::Cursor;
-use std::sync::atomic::{self, AtomicUsize, AtomicPtr};
-use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
+use core::{cmp, fmt, mem, hash, ops, slice, ptr, usize};
+use core::borrow::Borrow;
+use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
+use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
+
+#[cfg(feature = "std")]
```
alexcrichton: Did this file need to change? It looks like the whole module is only compiled when `std` is enabled?

alexcrichton:
One thing I've found useful is to have a module just called `std` which contains all std-relevant implementations of traits and such which is #[cfg]'d off by default. Would that be possible here to cut down on #[cfg] traffic?

src/bytes.rs
carllerche: Yeah, I'm pretty sure changes to this file could be reverted.

carllerche:
Sorry this has been left to stale. I will try to get back to this soon.

Kixunil:
What's the status of this?

tarcieri:
This PR, in and of itself, is pretty useless: as @alexcrichton noted in the line notes none of `bytes.rs` is available with this. My changes there were left over from me splitting an original PR into two parts, the other being #153, where I had parts of `bytes.rs` working when obtaining heap-allocated data structures through unstable, nightly-only allocator APIs. Without an allocator none of `bytes.rs` can be used.

Even if this PR and #153 were merged (and it looks like #153 won't ever be merged) there's a bigger problem though: `std::io::Cursor` is a pretty essential part of how `bytes`'s API works presently.

For those reasons I've created a mostly-compatible fork of bytes here, which vendors `std::io::Cursor`:

https://github.com/microcrates/bytes

My understanding is bytes 0.5 will redo the API so as not to need `std::io::Cursor`, and [future work on std should allow stable access to allocators](http://aturon.github.io/2018/02/06/portability-vision/) in what are presently considered "no_std" environments (i.e. no_std-with-a-heap environments). When these things happen `bytes` should hopefully just work in these environments.

Kixunil:
To clarify this, I'm really only interested in the traits and their most basic methods. I'd find it more elegant if they were split, but I doubt I could convince enough people.

carllerche:
This has been inactive. Unless there is interest to champion this over the finish line, it probably won't get more attention.

As such, I'm going to close this. A new PR can be opened once there is time to work on this.

Kixunil:
@carllerche How would you feel about putting the traits into a separate `no_std`- friendly crate? I might find some time for that...

carllerche:
@Kixunil I have no problem moving the traits, but they fundamentally depend on `io::Cursor` etc... for now. I do not think that progress can be made w/o a breaking change, and there is no plan to issue one immediately.

It might be possible to make progress w/o a huge breaking change, but doing so would be pretty complicated and not something that I have time for currently.

tarcieri:
Yeah, fixing #75 is still a blocker on any `no_std` support

PR #138, opened 2017-06-20
Fix index-oob panic in Take::bytes

danburkert:
The panic happens when `inner.bytes()` returns a slice smaller than the
limit.


carllerche:
grrr... yes! Thanks!

Could I trouble you for a unit test to avoid future regression? :)

danburkert:
Yep, can do soon.

danburkert:
Basic test case added.

PR #141, opened 2017-06-30
Adding convenience PartialEq impl for Bytes and Bytesmut

SpamapS:


carllerche:
Thanks! Looks good to me!

PR #145, opened 2017-07-03
small fixups in bytes.rs

danburkert:
A few commits making independent simplifications to bytes.rs.  A lot of these are possible due to bumping the minimum `rustc` version.  I have a suspicion that the first commit is fixing some unintended UB, but I haven't been able to find any documentation to prove it.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1630,10 +1630,9 @@ impl Inner {
         if capacity <= INLINE_CAP {
             unsafe {
                 // Using uninitialized memory is ~30% faster
-                Inner {
-                    arc: AtomicPtr::new(KIND_INLINE as *mut Shared),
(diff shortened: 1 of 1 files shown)


src/bytes.rs
```diff
@@ -1725,8 +1694,8 @@ impl Inner {
     #[inline]
     fn set_inline_len(&mut self, len: usize) {
         debug_assert!(len <= INLINE_CAP);
-        let p: &mut usize = unsafe { mem::transmute(&mut self.arc) };
-        *p = (*p & !INLINE_LEN_MASK) | (len << INLINE_LEN_OFFSET);
+        let p: &mut _ = self.arc.get_mut();
```
carllerche: Super late... but why is the `&mut _` needed here?

PR #146, opened 2017-07-03
Vendor std::io::Cursor as buf::Cursor (closes #75)

### LINKED ISSUE #75: `io::Cursor` is a bit out of place

tarcieri:
Extracts the minimal parts of std::io::Cursor needed to support bytes use cases.

This is needed to support no_std as std::io is unavailable in these contexts.
It also (arguably) improves the API ergonomics by simplifying imports.

Per notes on #75, changes the u64 position to a u32 to save space.


src/buf/cursor.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn get_mut(&mut self) -> &mut T { &mut self.inner }
+
+    /// Returns the current position of this cursor.
+    pub fn position(&self) -> u32 { self.pos }
+
+    /// Sets the position of this cursor.
+    pub fn set_position(&mut self, pos: u32) { self.pos = pos; }
+}
+
+impl<T> Read for Cursor<T> where T: AsRef<[u8]> {
```
tarcieri: The `Read` and `BufRead` impls here aren't actually used by `bytes` at all. I have included them as without them the following example fails:

https://github.com/tarcieri/bytes/blob/4c3faa7e501ee683629b1792d1e1a290fc83ed4e/src/buf/writer.rs#L67

That said, perhaps people are expecting them? (and possibly `Write` as well?)

src/buf/buf.rs
```diff
@@ -15,8 +15,7 @@ use std::{cmp, io, ptr};
 /// The simplest `Buf` is a `Cursor` wrapping a `[u8]`.
 ///
 /// ```
-/// use bytes::Buf;
-/// use std::io::Cursor;
+/// use bytes::{Buf, Cursor};
```
tarcieri: I think this is a nice ergonomic improvement over:

```rust
use bytes::Buf;
use std::io::Cursor;
```

src/buf/cursor.rs
```diff
@@ -0,0 +1,69 @@
...
+/// implement [`Read`] and/or [`Write`], allowing these buffers to be used
+/// anywhere you might use a reader or writer that does actual I/O.
+///
+/// The standard library implements some I/O traits on various types which
+/// are commonly used as a buffer, like `Cursor<`[`Vec`]`<u8>>` and
+/// `Cursor<`[`&[u8]`][bytes]`>`.
+#[derive(Clone, Debug)]
+pub struct Cursor<T> {
+    inner: T,
+    pos: u32,
```
tarcieri: This is a `u64` in `std::io`, however @alexcrichton suggested using a smaller type here: https://github.com/carllerche/bytes/issues/75#issue-214793221

tarcieri:
Note this is a breaking API change which means it can't be landed until bytes 0.5.

alexcrichton:
In the technical sense this is unfortunately still a breaking change because `std::io::Cursor` no longer implements `Buf` and if you have code consuming `std::io::Cursor` created from `bytes` then it'll get a different `Cursor` now which'll break.

Whether it's actually breaking in a practical sense is a different question, though, which I'm not sure about.

tarcieri:
Yeah, duly noted this is a breaking change. I'm also uncertain it's the best path forward.

tarcieri:
After talking with @carllerche it sounds like if a breaking change is to happen here, it would probably be better to get rid of cursors altogether.

PR #147, opened 2017-07-04
Add optional `HeapSizeOf` impl

Vurich:
This is questionably useful for a shared buffer, since (for example) the sum of `heap_size_of` for a vector of `Bytes` values is going to be an upper bound on the actual amount of space used, but it's useful for our usecase nonetheless (we're replacing `Arc<Vec<u8>>` with `Bytes`, so the unreliability of `heap_size_of` is not a regression) and it might be helpful upstream.

Also, I know that this crate is full of subtle opportunities for unsafety and incorrect behaviour, so I hope by submitting a PR I might get any of those pointed out.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1405,6 +1405,12 @@ impl ops::Deref for BytesMut {
     }
 }
 
+impl AsMut<[u8]> for BytesMut {
+    fn as_mut(&mut self) -> &mut [u8] {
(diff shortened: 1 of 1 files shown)


Vurich:
I'm going to reopen this pulling off a feature branch so I can merge this and https://github.com/carllerche/bytes/pull/148 on my personal master (I need both for the work I'm doing)

PR #149, opened 2017-07-05
Add optional `HeapSizeOf` impl

Vurich:
This is questionably useful for a shared buffer, since (for example) the `heap_size_of` calculation for a vector of `Bytes` values is going to be an upper bound on the actual amount of space used, but it's useful for our usecase nonetheless (we're replacing `Arc<Vec<u8>>` with `Bytes`, so the unreliability of `heap_size_of` is not a regression) and it might be helpful upstream.

Also, I know that this crate is full of subtle opportunities for unsafety and incorrect behaviour, so I hope by submitting a PR I might get any of those pointed out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ categories = ["network-programming", "data-structures"]
 byteorder = "1.0.0"
 iovec = "0.1"
 serde = { version = "1.0", optional = true }
+heapsize = { version = "0.4.0", optional = true }
 
 [dev-dependencies]
 serde_test = "1.0"
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -363,6 +363,40 @@ const INLINE_CAP: usize = 4 * 8 - 1;
 #[cfg(target_pointer_width = "32")]
 const INLINE_CAP: usize = 4 * 4 - 1;
 
+// Optional Serde support
+#[cfg(feature = "heapsize")]
+mod heapsizeimpl {
+    extern crate heapsize;
+
+    use self::heapsize::HeapSizeOf;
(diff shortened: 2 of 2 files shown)


Vurich:
@carllerche Any update on this?

carllerche:
Thanks for the suggestion.

I am not super familiar with the heap size crate. However, it looks like it isn't at 1.0. I'm also not sure why this would be needed over checking the capacity of the value?

Vurich:
For generic contexts, mainly. This trait is used in Servo and Parity to allow memory limits and other stuff. I would personally prefer to use an arena with failable allocation but that's not available in Rust right now so we use this.

carllerche:
I'm leaning towards not adding a dependency because it doesn't seem to have reached a stable (1.0) point, it doesn't seem to have a ton of community adoption, and you can always wrap a Bytes value or take `Bytes::capacity()` as a good estimation.

I wonder if @alexcrichton has an opinion.

alexcrichton:
I'd personally feel the same as @carllerche, this is effectively adding a new public dependency which can make it difficult for pre-1.0 crates.

Vurich:
Unless I'm misunderstanding, it's only a dependency if you explicitly opt-in using the feature, right? I thought that optional dependencies aren't even resolved or downloaded unless the feature is enabled.

alexcrichton:
It's off by default, yes, but it's unfortunately still a public dependency of this crate. All that really means though is that mio can't upgrade the dependency on HeapSizeOf unless it also has a major version bump.

Vurich:
@alexcrichton Ah, I see. How irritating.

PR #150, opened 2017-07-06
Add conversion from `Bytes` to `Vec`

Vurich:
*Disclaimer: I am not certain this is safe. This needs checking.*

As far as I can tell, if the `Bytes` value is backed by a vec it's safe to take ownership of its data. There are other functions that do `Vec::from_raw_parts` on the data without locking. I'm confident about that, but the `Arc`-backed case I am less sure about. This needs checking but someone who knows how this works. I've tried to add tests that would fail in the case of memory unsafety (and it's not the kind of unsafety that would be optimised out by LLVM, so I'm not worried about that) but it's possible that I've misunderstood how this works.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1474,6 +1474,33 @@ impl From<Bytes> for BytesMut {
     }
 }
 
+impl From<Bytes> for Vec<u8> {
+    fn from(src: Bytes) -> Vec<u8> {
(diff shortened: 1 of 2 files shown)


Vurich:
Closed in favour of https://github.com/carllerche/bytes/pull/151

PR #153, opened 2017-07-18
"allocator" and "nightly" features (for no_std environments)

tonychain:
This gates all allocator-dependent features on an "allocator" feature.

It also adds a "nightly" feature which enables using the "allocator" feature in no_std environments. This requires using `#[feature(alloc)]` which requires nightly.

The "allocator" feature is automatically enabled when either the "std" or "nightly" features are enabled.

Travis CI is configured to check that builds succeed with both the "nightly" feature along with "std" and "nightly" in combination. To avoid the problem of nightly changes breaking the build, these combinations are specifically flagged as allowed failures in the Travis CI configuration.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,8 +70,13 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.4")]
+#![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate byteorder;
+
+#[cfg(feature = "std")]
+extern crate core;
+#[cfg(feature = "std")]
 extern crate iovec;
 
 pub mod buf;
@@ -88,8 +93,12 @@ pub use buf::{
     Take,
 };
 
+#[cfg(feature = "std")]
 mod bytes;
+#[cfg(feature = "std")]
 mod debug;
+
+#[cfg(feature = "std")]
 pub use bytes::{Bytes, BytesMut};
 
 pub use byteorder::{ByteOrder, BigEndian, LittleEndian};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,9 +71,12 @@
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.4")]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(feature = "nightly", feature(alloc))]
 
 extern crate byteorder;
 
+#[cfg(feature = "nightly")]
+extern crate alloc;
 #[cfg(feature = "std")]
 extern crate core;
 #[cfg(feature = "std")]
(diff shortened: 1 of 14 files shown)


tonychain:
Note this PR includes a commit from the no_std PR, which should get merged first:

https://github.com/carllerche/bytes/pull/135

src/lib.rs
```diff
@@ -70,8 +70,16 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.4")]
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(feature = "nightly", feature(alloc))]
```
tarcieri: @alexcrichton so really this is the part that requires nightly, and that's because even if I have an `#[allocator]` defined `Box`, `Vec`, and `String` do not get added to the prelude.

If they did, then we could simply have an enabled-by-default, non-nightly-dependent `alloc` cargo feature  for gating features that depend on `Box`, `Vec`, and `String`.

I'm not sure how hard that would be, but I imagine it would be immensely useful for `no_std` users who have a heap available.

src/lib.rs
```diff
@@ -98,3 +110,15 @@ pub use byteorder::{ByteOrder, BigEndian, LittleEndian};
 #[cfg(feature = "serde")]
 #[doc(hidden)]
 pub mod serde;
+
+/// Custom (internal-only) prelude for this module
```
tonychain: I made this to avoid having to individually import each of these types everywhere they're used.

Seems like a bit of a hack, but dramatically reduces the scope of what the `nightly` feature does which I think is nice.

It also makes this easy to change in the event the [liballoc/libcollections merge is reverted](https://github.com/rust-lang/rust/issues/43112)

jbowens:
Hey @carllerche. I'd love to see `no_std` for bytes. Is this pull request still the right direction?

carllerche:
Sorry to have left this without a response.

I don't think we want to add a dependency (even hidden behind a feature flag) on the allocator crate as it is highly unstable.

#135 is fine as an approach to support `no_std` but it does not include any types that require allocation.

I'm going to close this for now as #135 is the preferred path forward.

Ericson2314:
So the argument behind https://github.com/rust-lang/rfcs/pull/2480 is that while `alloc` is itself unstable, the items do not change much as std basically reexports them. So in practice is the a maintenance burden of adding optional features like this really that bad?

tarcieri:
If https://github.com/rust-lang/rfcs/pull/2480 lands, this will all work on stable. At that time I will happily revive this PR.

Ericson2314:
@tarcieri I am commenting here and there because I *both* want this PR and ones like it really bad *and* don't want to stabilize `alloc` just yet, and why I see no conflict between those two positions.

carllerche:
I have seen the stable build break due to adding nightly only features guarded with a feature flag.

PR #156, opened 2017-08-10
Update to byteorder v1.1.0

Thomasdezeeuw:


carllerche:
Shouldn't cargo automatically pull in the newest?

Thomasdezeeuw:
Probably however I couldn't use byteorder v1.1.0 as a direct dependency because Cargo was complaining about `bytes` using v1.0.0. I don't have the exact error message anymore, but I can try to get it again.

Thomasdezeeuw:
Apparently removin `Cargo.lock` fixes it, not sure why I didn't think about that last night. But in any case this isn't needed anymore. Sorry for the noise.

PR #162, opened 2017-11-03
Add support for unsplit() to BytesMut

jq-rs:
Add support for unsplit() to BytesMut which combines splitted contiguous memory blocks efficiently.

### DIFF
--- a/tests/test_bytes.rs
+++ b/tests/test_bytes.rs
@@ -544,7 +544,7 @@ fn unsplit_inline_arc() {
 
 #[test]
 fn unsplit_arc_inline() {
-    let mut buf = BytesMut::with_capacity(64); // arc
+    let mut buf = BytesMut::with_capacity(64);
     buf.extend_from_slice(b"aaaabbbbeeee");
 
     buf.split_off(8); //arc
@@ -573,15 +573,15 @@ fn unsplit_both_inline() {
 
 #[test]
 fn unsplit_arc_different() {
-    let mut buf = BytesMut::with_capacity(64); //arc
+    let mut buf = BytesMut::with_capacity(64);
     buf.extend_from_slice(b"aaaabbbbeeee");
 
-    buf.split_off(8);
+    buf.split_off(8); //arc
 
-    let mut buf2 = BytesMut::with_capacity(64); //arc
+    let mut buf2 = BytesMut::with_capacity(64);
     buf2.extend_from_slice(b"ccccddddeeee");
 
-    buf2.split_off(8);
+    buf2.split_off(8); //arc
 
     buf.unsplit(buf2);
     assert_eq!(b"aaaabbbbccccdddd", &buf[..]);
@@ -589,15 +589,12 @@ fn unsplit_arc_different() {
 
 #[test]
 fn unsplit_arc_non_contiguous() {
-    let mut buf = BytesMut::with_capacity(64); //arc
-    buf.extend_from_slice(b"aaaabbbbeeee");
+    let mut buf = BytesMut::with_capacity(64);
+    buf.extend_from_slice(b"aaaabbbbeeeeccccdddd");
 
-    buf.split_off(8);
+    let mut buf2 = buf.split_off(8); //arc
 
-    let mut buf2 = BytesMut::with_capacity(64); //arc
-    buf2.extend_from_slice(b"eeeeccccdddd");
-
-    let buf3 = buf2.split_off(4);
+    let buf3 = buf2.split_off(4); //arc
 
     buf.unsplit(buf3);
     assert_eq!(b"aaaabbbbccccdddd", &buf[..]);
@@ -605,11 +602,11 @@ fn unsplit_arc_non_contiguous() {
 
 #[test]
 fn unsplit_two_split_offs() {
-    let mut buf = BytesMut::with_capacity(64); //arc
+    let mut buf = BytesMut::with_capacity(64);
     buf.extend_from_slice(b"aaaabbbbccccdddd");
 
     let mut buf2 = buf.split_off(8); //arc
-    let buf3 = buf2.split_off(4);//arc
+    let buf3 = buf2.split_off(4); //arc
 
     buf2.unsplit(buf3);
     buf.unsplit(buf2);


src/bytes.rs
```diff
@@ -1324,6 +1324,45 @@ impl BytesMut {
...
+            ptr = self.inner.ptr.offset(self.inner.len as isize); 
+        }
+        if ptr == other.inner.ptr &&
+           self.inner.kind() == KIND_ARC &&
+           other.inner.kind() == KIND_ARC
+        {
+            // Contiguous blocks, just combine directly
+            self.inner.len += other.inner.len;
+            self.inner.cap += other.inner.cap;
+            drop(other);
```
asomers: This `drop` is redundant.  It will happen automatically.

src/bytes.rs
```diff
@@ -1324,6 +1324,45 @@ impl BytesMut {
...
+        let ptr;
+
+        unsafe {
+            ptr = self.inner.ptr.offset(self.inner.len as isize); 
+        }
+        if ptr == other.inner.ptr &&
+           self.inner.kind() == KIND_ARC &&
+           other.inner.kind() == KIND_ARC
+        {
+            // Contiguous blocks, just combine directly
```
asomers: How about a `debug_assert_eq!(self.arc, other.arc)`?

jq-rs:
Thanks for the suggestions, glad to hear there is a need for this new feature! As extend() should handle the other corner cases, I am not sure would these additional tests be useful enough for unsplit(). Drop is indeed redundant, just wanted to be explicit. I am not sure what would debug_assert_eq!(self.arc, other.arc) achieve?

Anyway, I am open to any changes that would get this merged, there could be also something else I have missed. What do you think @carllerche?

jq-rs:
I thought about it and debug_assert_eq!(self.arc, other.arc) would guarantee that reference count is always equal. Makes sense. I think the other corner case tests would need to be separate unit tests, not documentation tests. I'll proceed with the changes hopefully soon.

tests/test_bytes.rs
```diff
@@ -514,3 +514,103 @@ fn partial_eq_bytesmut() {
...
+
+#[test]
+fn unsplit_inline_arc() {
+    let mut buf = BytesMut::with_capacity(8); //inline
+    buf.extend_from_slice(b"aaaabbbb");
+    
+    let mut buf2 = BytesMut::with_capacity(64); 
+    buf2.extend_from_slice(b"ccccddddeeee");
+    
+    let _buf3 = buf2.split_off(8); //arc
```
asomers: What's the purpose of `_buf3`?

tests/test_bytes.rs
```diff
@@ -514,3 +514,103 @@ fn partial_eq_bytesmut() {
...
+    buf.unsplit(buf2);
+    assert_eq!(b"aaaabbbbccccdddd", &buf[..]);
+}
+
+#[test]
+fn unsplit_arc_inline() {
+    let mut buf = BytesMut::with_capacity(64); // arc
+    buf.extend_from_slice(b"aaaabbbbeeee");
+
+    let _buf2 = buf.split_off(8); //arc
```
asomers: What's the purpose of `_buf2`?

tests/test_bytes.rs
```diff
@@ -514,3 +514,103 @@ fn partial_eq_bytesmut() {
...
+fn unsplit_arc_different() {
+    let mut buf = BytesMut::with_capacity(64); //arc
+    buf.extend_from_slice(b"aaaabbbbeeee");
+
+    let _buf2 = buf.split_off(8);
+
+    let mut buf3 = BytesMut::with_capacity(64); //arc
+    buf3.extend_from_slice(b"ccccddddeeee");
+
+    let _buf4 = buf3.split_off(8);
```
asomers: I guess `_buf2` is there just to force `extend_from_slice` to do a reallocation.  But what about `_buf4`?

tests/test_bytes.rs
```diff
@@ -514,3 +514,103 @@ fn partial_eq_bytesmut() {
...
+    buf.extend_from_slice(b"aaaabbbbeeee");
+
+    let _buf2 = buf.split_off(8);
+
+    let mut buf3 = BytesMut::with_capacity(64); //arc
+    buf3.extend_from_slice(b"eeeeccccdddd");
+
+    let buf4 = buf3.split_off(4);
+    
+    buf.unsplit(buf4);
```
asomers: In this test `buf` and `buf4` have totally different arcs.  I think it would be a better to test if they shared the same arc, but noncontiguous portions.  Something like:
``` rust
buf.extend_from_slice(b"aaaabbbbeeeeffff");
let buf2 = buf.split_off(8);
let buf4 = buf2.split_off(8);
buf.unsplit(buf4);
```

tests/test_bytes.rs
```diff
@@ -514,3 +514,103 @@ fn partial_eq_bytesmut() {
...
+    assert_eq!(b"aaaabbbbccccdddd", &buf[..]);
+}
+
+#[test]
+fn unsplit_three_arcs() {
```
asomers: Why do you call this "three_arcs"?  Don't all three buffers share the same arc?  And what's the purpose of `_buf3`?

tests/test_bytes.rs
jq-rs: A good point, it does not really make the test any clearer for the reader. Now removed all of them.

tests/test_bytes.rs
jq-rs: Now renamed to two_split_offs.

tests/test_bytes.rs
```diff
@@ -514,3 +514,104 @@ fn partial_eq_bytesmut() {
...
+    buf.extend_from_slice(b"aaaabbbbeeee");
+
+    buf.split_off(8);
+
+    let mut buf2 = BytesMut::with_capacity(64); //arc
+    buf2.extend_from_slice(b"eeeeccccdddd");
+
+    let buf3 = buf2.split_off(4);
+
+    buf.unsplit(buf3);
```
asomers: buf and buf3 still have independent arcs.  I think it would make a better test if they were noncontiguous parts of the same arc.

jq-rs: Yes, it would make a better test as we have own test for separate arcs. Thanks, now fixed with 288efed

asomers:
ping.  Could you please squash these commits?  Then maybe @carllerche can review it too.

jq-rs:
They can be squashed when PR is merged. Or do you mean some manual squash perhaps?

carllerche:
Thanks! Looks good. I can squash when merging. I prefer keeping the commits unsquashed in the PR :)

carllerche:
There are actually a few further improvements that could be made if you want to try:

* Check if `other` is empty, in which case, no-op.
* Check if `self` is empty, in which case, just set `*self = other`.
* Use `extend_from_slice` instead of `extend` as it is more optimized for slices instead of iterators.

PR #165, opened 2017-11-14
add support for generating random sequences with QuickCheck

### LINKED ISSUE #103: Implement quickcheck::Arbitrary for Bytes and BytesMut
See http://burntsushi.net/rustdoc/quickcheck/trait.Arbitrary.html for a description.

Quickcheck is a randomized testing tool that lets you check general properties. To make your type part of the Quickcheck ecosystem, you need to implement Arbitrary, which involves two methods:

(1) generate a random input
(2) shrink a failing input

This should be relatively easy -- probably just using the `Vec` code underneath.

This can be behind a feature gate to make sure that if you aren't using quickcheck otherwise, you don't have to pull it in.

sunshowers:
Closes #103.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,9 @@ matrix:
     # Serde implementation
     - env: EXTRA_ARGS="--features serde"
 
+    # QuickCheck implementation
+    - env: EXTRA_ARGS="--features quickcheck"
+
 before_install: set -e
 
 install:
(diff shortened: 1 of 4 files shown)


sid0:
Not sure why it didn't get updated, but the Travis build is complete.

sunshowers:
Probably no longer useful :)

Issue #74, opened 2017-03-16
Bytes/BytesMut vs Buf/BufMut is confusing in terms of naming


alexcrichton:
I'd expect `Buf for Bytes` and `BufMut for BytesMut` in terms of what the name implies, but unfortunately there's no `Buf for Bytes`

danburkert:
I'd like to make the case for providing an `impl Buf for Bytes`.

I also expected the `BytesMut` : `BufMut`, `Bytes` : `Buf` symmetry to hold when I first started working on with the crate.  For my usecase of `Bytes`, which is decoding data structures, it would be extremely handy to have the `Buf` implementation built-in to `Bytes`, instead of having to wrap in a `Cursor`.

Towards that end, I put together a [proof-of-concept](https://github.com/danburkert/bytes/commit/6617157bda5131e0acc7bf331a0f76cccfb9593e) showing that `Buf` can be implemented for `Bytes` using nothing but the current public API.  The PoC is not meant to be an efficient implementation, but instead to prove that it can be done.

cc @carllerche

carllerche:
I had some thoughts on how to impl it while avoiding the allocation. I'll try to work on it.

However, the changes of types for the associate types of trait impls can't change until 0.5 :(

danburkert:
> However, the changes of types for the associate types of trait impls can't change until 0.5 :(

danburkert:
Has reserving the first 4 or 8 bytes of the allocated buffer for the reference count been considered?  This would 'waste' 4 or 8 bytes in the non-shared case, but it cuts down an allocation in the shared case.  I think it would also make efficiently impl'ing `Buf` easier.

carllerche:
#166 provides the `advance` function for `Bytes` and `BytesMut`. This unblocks implementing `Buf` for these two types. The actual implementation of `Buf` has to be done in 0.5.

carllerche:
Is this still true if `Bytes` impls `Buf` and `BytesMut` impls `BufMut`?

xoac:
Shouldn't this be closed? In 0.5.x:
- `Buf` is implemented for `Bytes` and for `BytesMut`.
- `BufMut` is implemented for `BytesMut`

PR #166, opened 2017-11-16
Add `advance` on `Bytes` and `BytesMut`

carllerche:
In order to avoid unnecessary allocations, a `Bytes` structure remembers
the capacity with which it was first created. When a reserve operation
is issued, this original capacity value is used to as a baseline for
reallocating new storage.

Previously, this original capacity value was stored in its raw form. In
other words, the original capacity `usize` was stored as is. In order to
reclaim some `Bytes` internal storage space for implementing `advance`,
this original capacity value is compressed from requiring 16 bits to 3.

To do this, instead of storing the exact original capacity. The original
capacity is rounded down to the nearest power of two. If the original
capacity is less than 1024, then it is rounded down to zero. This
roughly means that the original capacity is now stored as a table:

```
0 => 0
1 => 1k
2 => 2k
3 => 4k
4 => 8k
5 => 16k
6 => 32k
7 => 64k
```

For the purposes that the original capacity feature was introduced, this
is sufficient granularity.

The second commit implements `advance` by using this new storage capacity
to store the window offset into the vec. The `ptr`, `len`, and `cap` fields
represent the active window, i.e. the user view after a call to `advance`. This
makes common operations like dereferencing the slice fast.

Relates to #74

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -343,6 +343,14 @@ const MIN_ORIGINAL_CAPACITY_WIDTH: usize = 10;
 const ORIGINAL_CAPACITY_MASK: usize = 0b11100;
 const ORIGINAL_CAPACITY_OFFSET: usize = 2;
 
+// When the storage is in the `Vec` representation, the pointer can be advanced
+// at most this value. This is due to the amount of storage available to track
+// the offset is usize - number of KIND bits and number of ORIGINAL_CAPACITY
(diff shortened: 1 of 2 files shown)


alexcrichton:
Sorry I'm pretty out of touch with the implemenation here, you'll likely wish to find a different reviewer

Issue #172, opened 2018-01-03
Improve unsplit


carllerche:
A continuation of #162. Specifically:

There are actually a few further improvements that could be made [to unsplit]:

* Check if `other` is empty, in which case, no-op.
* Check if `self` is empty, in which case, just set `*self = other`.
* Use `extend_from_slice` instead of `extend` as it is more optimized for slices instead of iterators.

jq-rs:
Worthy improvements, I'll take a look soon unless someone gets them in first.

jq-rs:
Thanks for the merge! I tested the unsplit() version with my application and seems to function very well. If you could release a new bytes version to crates.io with unsplit() support, that would be just great. Thanks!

carllerche:
Could you file an issue requesting a new release?

carllerche:
Closed by #173

PR #173, opened 2018-01-05
Unsplit improvements

jq-rs:
Handle empty self and other for unsplit. Change extend() to extend_from_slice().

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -295,7 +295,8 @@ pub struct BytesMut {
 #[cfg(target_endian = "little")]
 #[repr(C)]
 struct Inner {
-    // WARNING: Do not access the fields directly unless you know what you are doing. Instead, use the fns. See implementation comment above.
+    // WARNING: Do not access the fields directly unless you know what you are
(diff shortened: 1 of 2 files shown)


jq-rs:
Fixes #172

src/bytes.rs
```diff
@@ -1401,6 +1401,15 @@ impl BytesMut {
     pub fn unsplit(&mut self, other: BytesMut) {
         let ptr;
 
+        if other.inner.len == 0 {
```
carllerche: You cannot access the fields directly as they may be used to store the inline data. Instead, use the fns.

src/bytes.rs
```diff
@@ -1401,6 +1401,15 @@ impl BytesMut {
...
+            return;
+        }
+
+        if self.inner.len == 0 {
```
carllerche: Same here.

jq-rs:
Thanks for the comments! Yep, an fn should be used of course. Somehow I already forgot about the inline. Now tests seem to pass just fine.

PR #176, opened 2018-01-11
Optionally disable inlining

### LINKED ISSUE #175: Optionally disable inlining
Inlining a `Bytes` or `BytesMut` is undesirable when one needs to pass a raw pointer to an asynchronous C API.  If the `Bytes` is stack-allocated and inline, the location of the slice could change leaving the C library with a dangling pointer.  One solution to this problem is to always allocate your `Bytes` within a `Box`, but that leads to unnecessary malloc work.  A better solution would be to have a way to disable inline allocation.  I propose adding two methods to support this:

- `Bytes::is_inline`.  This method already exists, we would just need to make it public.
- `Bytes::with_capacity_stable`.  This would be like `with_capacity`, but it would guarantee stable storage.  An object initialized with this method would always return false for `is_inline`.

asomers:
Add with_capacity_stable constructors that will always use stable
storage rather than inlined storage.  They are useful when working with
asynchronous C APIs that take raw pointers.

Fixes #175

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -417,6 +417,27 @@ impl Bytes {
         }
     }
 
+    /// Creates a new `Bytes` with the specified capacity, using stable storage
+    ///
+    /// Exactly like `with_capacity`, but never uses the inline optimization.
+    /// The resulting `Bytes` can be used with asynchronous C APIs that require
+    /// stable storage.
+    ///
+    /// # Examples
+    /// ```
+    /// use bytes::Bytes;
+    ///
+    /// let mut bytes = Bytes::with_capacity_stable(4);
+    ///
+    /// assert!(!bytes.is_inline());
+    /// ```
+    #[inline]
+    pub fn with_capacity_stable(capacity: usize) -> Bytes {
+        Bytes {
+            inner: Inner::with_capacity_stable(capacity),
+        }
+    }
+
     /// Creates a new empty `Bytes`.
     ///
     /// This will not allocate and the returned `Bytes` handle will be empty.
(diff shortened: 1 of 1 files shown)


src/bytes.rs
```diff
@@ -482,6 +503,19 @@ impl Bytes {
...
+    /// Return true if the `Bytes` uses inline allocation
+    ///
+    /// # Examples
+    /// ```
+    /// use bytes::Bytes;
+    ///
+    /// assert!(Bytes::with_capacity(4).is_inline());
+    /// assert!(!Bytes::with_capacity(1024).is_inline());
+    /// ```
+    pub fn is_inline(&self) -> bool {
```
carllerche: These functions are good to expose.

src/bytes.rs
```diff
@@ -417,6 +417,27 @@ impl Bytes {
...
+    /// # Examples
+    /// ```
+    /// use bytes::Bytes;
+    ///
+    /// let mut bytes = Bytes::with_capacity_stable(4);
+    ///
+    /// assert!(!bytes.is_inline());
+    /// ```
+    #[inline]
+    pub fn with_capacity_stable(capacity: usize) -> Bytes {
```
carllerche: This capability is already available by converting a vec. So, you would do: `Bytes::from(Vec::with_capacity(8))`. Do you think adding this constructor is still necessary?

Either way, the documentation should specifically call out that converting a Vec -> Bytes will keep the vec's storage intact.

asomers: I suppose that would work since the specification does guarantee that `Vec` will always use heap storage [1].  But it feels hacky to tell consumers to use such a roundabout means of construction.  Ultimately it's up to you of course.  Let me know if you'd like me to change it.

[1] https://doc.rust-lang.org/std/vec/struct.Vec.html#guarantees

carllerche: I personally tend to err on the side of adding less new APIs. New APIs can always be added, but removing them is trickier.

If it turns out to be a point of confusion, we can add add the fn at that point.

carllerche:
Thanks!

PR #197, opened 2018-05-04
Recycle space when reserving from Vec-backed Bytes

### LINKED ISSUE #193: BytesMut memory usage grows without bound
I'm working on a high-throughput network service with Tokio, and I'm running into an issue where the memory used by a `BytesMut` grows without bound as more and more messages are processed. After some debugging, I was able to come up with this minimal example:

    extern crate bytes;
    use bytes::*;

    fn leaks_memory() {
        let mut data = BytesMut::from(Vec::with_capacity(16));

        loop {
            data.reserve(4);
            data.put(&b"leak"[..]);
            data.advance(4);
        }
    }

    fn also_leaks_memory() {
        let mut data = BytesMut::with_capacity(8192);

        loop {
            data.reserve(4);
            data.put(&b"leak"[..]);
            data.advance(4);
        }
    }

    fn constant_memory() {
        let mut data = BytesMut::from(Vec::with_capacity(16));

        loop {
            data.reserve(4);
            data.put(&b"leak"[..]);

            data = data.split_off(4);
        }
    }

    fn also_constant_memory() {
        let mut data = BytesMut::with_capacity(16);

        loop {
            data.reserve(4);
            data.put(&b"leak"[..]);
            data.advance(4);
        }
    }

It looks like what's happening is that a `BytesMut` backed by a `Vec` isn't actually reusing space at the beginning of the buffer when `reserve` is called, but only if it was moved forwards with `advance`. I'm not sure whether any of the other methods (e.g. `truncate`) cause the problem. The documentation says `reserve` will try to recycle existing space (which it's clearly doing in `also_constant_memory`) but it seems to be failing when it shouldn't.

nzentzis:
BytesMut::reserve, when called on a BytesMut instance which is backed by a non-shared Vec<u8>, would previously just delegate to Vec::reserve regardless of the current location in the buffer. If the Bytes is actually the trailing component of a larger Vec, then the unused space won't be recycled. In applications which continually move the pointer forward to consume data as it comes in, this can cause the underlying buffer to get *extremely* large.

This just checks whether there's extra space at the start of the backing Vec in this case, and reuses the unused space if possible instead of allocating. Fixes #193.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -2258,7 +2258,9 @@ impl Inner {
             unsafe {
                 let (off, prev) = self.uncoordinated_get_vec_pos();
 
-                if off >= additional {
+                // Only reuse space if we stand to gain at least capacity/2
+                // bytes of space back
+                if off >= additional && off >= (self.cap / 2) {
                     // There's space - reuse it
                     //
                     // Just move the pointer back to the start after copying
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -2250,20 +2250,40 @@ impl Inner {
         }
 
         if kind == KIND_VEC {
-            // Currently backed by a vector, so just use `Vector::reserve`.
+            // If there's enough free space before the start of the buffer, then
+            // just copy the data backwards and reuse the already-allocated
+            // space.
+            //
+            // Otherwise, since backed by a vector, use `Vec::reserve`
             unsafe {
-                let (off, _) = self.uncoordinated_get_vec_pos();
-                let mut v = rebuild_vec(self.ptr, self.len, self.cap, off);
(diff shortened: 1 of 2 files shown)


carllerche:
The implementation looks good. I had two thoughts. First, would you be able to add tests or verify that existing tests hit these conditions? Second, I worry that unconditionally doing a memmove w/o allocating could get into pathological situations where you end up moving large amounts of memory a *lot*. Perhaps there should be some heuristic to determine if the existing buffer should be used vs. allocating a new one. Maybe you gain at least 50% capacity or something. Thoughts?

carllerche:
Thanks!

PR #203, opened 2018-05-24
Added a resize function for BytesMut

### LINKED ISSUE #202: Add a resize function
It is sometimes desirable to be able to have a zeroed buffer of a particular length, e.g. when using [`rand::RngCore::fill_bytes`](https://docs.rs/rand/0.5.0/rand/trait.RngCore.html#tymethod.fill_bytes).

Currently the only way I know how to do this with an existing `BytesMut` without using unsafe code is the following:

    fn generate_bytes(length: usize) -> BytesMut {
        let mut bytes = BytesMut::new();
        bytes.extend_from_slice(&vec![0; length][..]);

        let rng = ::rand::thread_rng();
        rng.fill_bytes(&mut bytes[..]);
        bytes
    }

This has the unfortunate side effect of allocating an intermediate `Vec`.

`vec![0; length].into()` may be used to create a new `BytesMut` of a specified length however doesn't help when using an existing `BytesMut` value.

I would like it if the following were possible:

    fn generate_bytes(length: usize) -> BytesMut {
        let mut bytes = BytesMut::new();
        bytes.resize(length, 0);

        let rng = ::rand::thread_rng();
        rng.fill_bytes(&mut bytes[..]);
        bytes
    }

This function would be equivalent to [`Vec::resize`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize) in behaviour.

Lukazoid:
Closes #202

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1279,8 +1279,8 @@ impl BytesMut {
     /// Resizes the buffer so that `len` is equal to `new_len`.
     /// 
     /// If `new_len` is greater than `len`, the buffer is extended by the 
-    /// difference with each additional slot zeroed. If `new_len` is less than
-    /// `len`, the buffer is simply truncated.
+    /// difference with each additional byte set to `value`. If `new_len` is 
+    /// less than `len`, the buffer is simply truncated.
     /// 
     /// #Examples
     /// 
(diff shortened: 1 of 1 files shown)


carllerche:
Thanks for this.

I rebased the branch on top of `v0.4.x` and tweaked the doc section style.

Issue #198, opened 2018-05-09
Want a way to go from a &[u8] sub-slice to a Bytes


federicomenaquintero:
I'm using Bytes to hold data and then keep around slices from it.  The "main" data gets parsed with nom, which often returns `&[u8]` slices.  It would be convenient for my code to be able to go from those slices to `Bytes` slices on the original data:

```rust
pub trait Subslice {
    fn subslice(&self, sub: &[u8]) -> Self;
}

impl Subslice for Bytes {
    fn subslice(&self, sub: &[u8]) -> Bytes {
        let bytes_p = self.as_ptr() as usize;
        let bytes_len = self.len();

        let sub_p = sub.as_ptr() as usize;
        let sub_len = sub.len();

        assert!(sub_p >= bytes_p);
        assert!(sub_p + sub_len <= bytes_p + bytes_len);

        let sub_offset = sub_p - bytes_p;

        self.slice(sub_offset, sub_offset + sub_len)
    }
}
```

Do you think this should be in the Bytes API?  I already have tests for the above if you'd like a PR :smile:

federicomenaquintero:
(To clarify: the `subslice` function would be in the `Bytes` impl, not in a separate trait.  I haven't thought about doing this on `BytesMut`.)

carllerche:
Generally, I try to stay on the side of less APIs vs .more if it is possible to impl the fn outside of the library w/ no unsafe code. To include an API, it should be generally useful.

I think something like this might fit the criteria.

Of course, then we have to bikeshed the name :)

federicomenaquintero:
`Bytes.slice_to_bytes`?  `Bytes.get_subslice`?  `Bytes.subslice`? :laughing:

PR #208, opened 2018-05-31
Add a subslice function for Bytes (#198)

federicomenaquintero:


carllerche:
Sorry for letting this hang. I will try to up prioritize it.

The implementation is good. The name is bugging me. I would like it to match the pattern of the others. So, I would like a name like:

`slice_???`

Except, I can't think of anything good :(

Best I could think of is `slice_subset`... but I Don't like that either. Thoughts?

carllerche:
Or, what about `slice_for`?

```rust
let bytes = Bytes::from(&b"012345678"[..]);
let as_slice = bytes.as_ref();
let sub = &as_slice[2..6];
let sub = bytes.slice_for(&sub);
assert_eq!(&sub[..], b"2345");
```

carllerche:
Another (via Ralith in gitter): `slice_ref`

```rust
let bytes = Bytes::from(&b"012345678"[..]);
let as_slice = bytes.as_ref();
let sub = &as_slice[2..6];
let sub = bytes.slice_for(&sub);
assert_eq!(&sub[..], b"2345");
```

carllerche:
I'm like `slice_ref` as it follows the `slice_` pattern and implies that you are slicing using a reference to self.

`bytes.slice_ref(self.as_ref()) == bytes`

federicomenaquintero:
Cool, I'll rebase this and use `slice_ref`.

federicomenaquintero:
I've renamed the function to `slice_ref()` and the `sub` argument to `subset`.  Also changed the test names to make it more clear what they are doing.

carllerche:
Thanks for sticking with it 👍

carllerche:
I'm going to rebase the PR against the `0.4.x` branch.

carllerche:
CI hung for some reason...

federicomenaquintero:
Yay, thanks!

Issue #158, opened 2017-09-28
`BytesMut` should impl IntoBuf


carllerche:
See: https://github.com/tokio-rs/tokio-io/issues/72

seanmonstar:
It looks like this is already the case in 0.4.x. Should it still return a `Cursor<BytesMut>`? Or should `BytesMut` implement `Buf` directly instead?

carllerche:
`BytesMut` should impl `Buf` directly. Maybe it does already? Either way, it has the capability to do so.

YetAnotherMinion:
I ran into problems with `impl Buf for BytesMut`. Specifically when removing `advance` from `BytesMut` and using the `Buf`'s impl reveals that there are conflicting `take` methods on `Buf` and `BytesMut`.

`BytesMut::take` takes everything and passes ownership into a new `BytesBuf`, while `Buf::take` allows configuring how much is taken, but during the transferring ownership instead portion of its act, it consumes `self`, basically acting like truncate (without changing the underlying storage).

Can someone here give me more background information about the use case for the two different functions? I do not think the functionality of the two can be merged because they are doing fundamentally different things with ownership.

 The simplest solution I can think of is just renaming the functions. I think that `BytesMut::take` should keep the name, and `Buf::take` should be renamed to something like `Buf::prefix` (and `Take` to `Prefix`).

mzabaluev:
I, too, was confused with the disparate `take` methods. But my solution for this would be to replace `BytesMut::take` with a [trait API](https://github.com/mzabaluev/strchunk/blob/e4b1260215a1496567a73450945679cb09dd080a/src/split.rs#L139) that would work on both `Bytes` and `BytesMut` (and their UTF-8 counterparts `StrChunk`/`StrChunkMut` being incubated in that crate) and allow the user to specify a range to take. So `received_bytes.take()` would be `received_bytes.take_range(..)`.

mzabaluev:
> So `received_bytes.take()` would be `received_bytes.take_range(..)`.

I have now distilled that into crate [range-split](https://crates.io/crates/range-split).

Issue #196, opened 2018-05-01
Consider adding `IntoBuf::len`


carllerche:
Without this function, it is not possible to have any insight into the length of the buffer that will be returned by `into_buf()`.

Issue #139, opened 2017-06-22
Change From<&'a [u8]> to an explicit function


seanmonstar:
Most of the `From` conversions to `Bytes` are shallow copies, just grabbing a pointer and pulling it in. There is also the `from_static` function to save copying a static set of bytes. However, the `From<&'a [u8]>` impl will copy the slice into a new buffer (either inline or a new heap allocation, depending on size). 

Unfortunately, that `&'a [u8]` means it applies to `&'static [u8]` as well. In cases where one is generic over `Into<Bytes>`, a user can easily and accidentally cause copies of static slices, instead of remembering to call `Bytes::from_static`. This issue proposes to change the defaults, so that **the faster way is easier**.

- Remove `From<&'a [u8]>` for `Bytes`/`BytesMut`.
- Add `From<&'static [u8]>` for `Bytes`/`BytesMut`.
- Add a new constructor that can take `&'a [u8]`, that explicitly shows a copy will happen.

The name of this new method could be a few things, these just some examples that occurred to me:

- `Bytes::copy(slice)`
- `Bytes::copy_from(slice)`

seanmonstar:
Oh, this would be a breaking change, so would require (block?) 0.5.

YetAnotherMinion:
Just to clarify, 

`From<&'a [u8]> for BytesMut` can stay because static slices are not mutable, so a copy has to happen from the  `'static`  slice when creating BytesMut.

So only Bytes would have its `From<&'a [u8]>` removed.

carllerche:
Mentioned by @arthurprs:

> While I understand the reasoning this severely limits Byte as a Vec replacement, like backing a https://crates.io/crates/string

Specifically, this issue: https://github.com/carllerche/string/issues/6

PR #210, opened 2018-06-01
[WIP] Implement everything for 0.5

YetAnotherMinion:
I am working off of the list of issues under 0.5 milestone.

This should address: #158, #75, #139, #196, #74 

## Changelog

- Renames the `Buf::take` method to `Buf::limit` to better describe what that method does. The rename was necessary to prevent ambiguity with `BytesMut::take`. I renamed `Take` to `Limit` to match the constructor method name.
-  Removed `Buf` impl for Cursor<Bytes> and replaced it with `impl Buf for Bytes`

-  Removed `Buf` impl for `Cursor<BytesMut>` and replaced it with `impl Buf for BytesMut`

- Renamed `Iter` to `IntoIter` to match naming convention of rust where iterators that move the collection are named `IntoIter`

- Adds `impl<T: Buf> IntoIterator for T` that constructs a `bytes::buf::IntoIter` struct.

- Removs `iter` method from `Buf` trait. The existing behavior with respect to ownership matched the convention of `into_iter` instead of `iter`. The builtin `IntoIterator` trait already provides for `into_iter`, so I decided to avoid ambiguity by just removing `Buf::iter` instead of a rename. It is up to each container implementing Buf to implement `iter` efficiently if possible.

- Adds `Bytes::iter` which returns `slice::Iter`.

- Adds `BytesMut::iter` which returns `slice::Iter`.

- Adds `Bytes::copy_from_slice(src: &'a [u8])` which constructs a new  

- Removed `impl<'a> From<&'a [u8]> for Bytes` and replaced with `impl From<&'static [u8]> for Bytes` which does not copy the data.

- Adds `IntoBuf::len` which returns the size of the buffer that would be created by calling `into_buf()` on `self`.


seanmonstar:
Thanks for taking this on!

- As a suggestion, I think changing to `Buf::limit(n)` may be clearer. I'm not a fan of `Iterator::take(n)`, since it sounds like an immediate action, but is actually lazy. `Buf::prefix(n)` might imply that `n` bytes are being prefixed before this buffer.
- There was a reason for `Buf::iter`, which was to provide an `Iterator` as long as you implement `Buf`. This could be helpful for composite buffer types. However, I agree with your decision. There isn't a need to provide an `Iter` implementation over `Buf`, it is more efficient for types to implement their own iterators if they can.

YetAnotherMinion:
@seanmonstar, you mentioned in
https://github.com/carllerche/bytes/issues/75#issuecomment-372767509 that Bytes and BytesMut should implement IntoIterator. My current patch completely removed IntoIterator impl on these types in favor of a blanket impl for `T: Buf`. 

Is this design acceptable?

YetAnotherMinion:
For reference, this PR fixes #158, #75

seanmonstar:
@YetAnotherMinion I think `Bytes` and `BytesMut` probably should implement `IntoIterator`, since it's needed to allow `a_bytes_mut.extend(some_other_bytes)`.

YetAnotherMinion:
@seanmonstar They do implement `IntoIterator` through their implementation of `Buf`.

Apply this patch to this PR branch to see it working.
```diff
diff --git a/tests/test_chain.rs b/tests/test_chain.rs
index 8877258..42a16eb 100644
--- a/tests/test_chain.rs
+++ b/tests/test_chain.rs
@@ -48,6 +48,15 @@ fn iterating_two_bufs() {
 }
 
 #[test]
+fn dirty_hack() {
+    let mut a = BytesMut::from(&b"hello"[..]);
+    let b = Bytes::from(&b"world"[..]);
+    a.extend(b);
+
+    assert_eq!(a.bytes(), &b"helloworld"[..]);
+}
+
+#[test]
 fn vectored_read() {
     let a = Cursor::new(Bytes::from(&b"hello"[..]));
     let b = Cursor::new(Bytes::from(&b"world"[..]));
```

seanmonstar:
Oh OK, I get it! That's probably good then.

YetAnotherMinion:
now also addresses #139

arthurprs:
> Removed impl<'a> From<&'a [u8]> for Bytes and replaced with impl From<&'static [u8]> for Bytes which does not copy the data.

Is this really a good idea?

arthurprs:
While I understand the reasoning this severely limits Byte as a Vec<u8> replacement, like backing a https://crates.io/crates/string

carllerche:
@YetAnotherMinion Sorry for the radio silence, it was not cool on my part to drop the ball. This PR is great work and very helpful.

The 0.5 release delay is mostly due to an attempt to coordinate a release with Tokio 0.2 release. Tokio includes `Bytes` as a public dependency in core crates like `tokio-io`, which means that users of Tokio would have to stay on 0.4.

0.5 represents a cleanup pass and, as such, doesn't really have a reason to ship on its own. Once Tokio 0.2 gets closer (which I hope to start tracking better), we can revive this PR.

seanmonstar:
Sorry for it taking so long, we're now actively working on 0.5 things. I've started splitting this PR into specific features, such as #260. I'll close once all have been merged.

YetAnotherMinion:
Hey, thanks for picking this stuff up again. I ended up using a private fork at work and moved on. It will be nice to switch back over coming soon. If there are still some other issues for 0.5 that need help, I am happy to pitch in.

seanmonstar:
We've released 0.5!

PR #214, opened 2018-06-15
Implement IntoBuf for mut slices.

espindola:
With this if foo is a mutable slice, it is possible to do

foo.into_buf().put_u32_le(42);

Before this patch into_buf would create a Cursor<&'a [u8]> and it
would not be possible to write into it.


espindola:
A friendly ping.

espindola:
I have changed the base to master and added a test.

PR #216, opened 2018-06-16
Clarify license as MIT

### LINKED ISSUE #215: Clarify license(s) this code is under
https://github.com/carllerche/bytes/blob/master/README.md#license currently says:

> bytes is primarily distributed under the terms of both the MIT license and the Apache License (Version 2.0), with portions covered by various BSD-like licenses.

What portions (if any) are covered by BSD-like licenses and which specific licenses cover these portions?

carllerche:
The intent of the license was to dual license MIT & Apache 2.0. However,
the messaging was copy / pasted from rust-lang.

Clarify the license as exclusively MIT.

Fixes #215


amyrrdakhan18-ux:
# 215

PR #217, opened 2018-06-25
optimize Inner::shallow_clone

seanmonstar:
- Clones when the kind is INLINE or STATIC are sped up by over double.
- Clones when the kind is ARC are spec up by about 10%. (Previously was 33%, but was using `ptr::copy` and TSAN warnings were triggered. While we *think* that would be safe because it's down only **after** an `Acquire` load, this PR opts to be conservative until we can verify it that it *certainly* is safe.)

Before

```
test clone_arc    ... bench:         27,175 ns/iter (+/- 1,195)
test clone_inline ... bench:         13,778 ns/iter (+/- 1,335)
test clone_static ... bench:         13,736 ns/iter (+/- 599)
```

After:

```
test clone_arc    ... bench:         25,398 ns/iter (+/- 1,209)
test clone_inline ... bench:          5,568 ns/iter (+/- 285)
test clone_static ... bench:          5,562 ns/iter (+/- 151)
```

## Warning

Modifies unsafe code, I could have goofed on something...

### DIFF
--- a/benches/bytes.rs
+++ b/benches/bytes.rs
@@ -114,6 +114,39 @@ fn deref_two(b: &mut Bencher) {
 }
 
 #[bench]
+fn clone_inline(b: &mut Bencher) {
+    let bytes = Bytes::from_static(b"hello world");
+
+    b.iter(|| {
+        for _ in 0..1024 {
+            test::black_box(&bytes.clone());
+        }
+    })
+}
+
+#[bench]
+fn clone_static(b: &mut Bencher) {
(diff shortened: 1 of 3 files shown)


seanmonstar:
Seems this triggers some TSAN warnings, since this is now calling `self.kind()` instead of `self.is_inline()` (which has a suppression configured). What's the preferred way to fix this? I can add a suppression to `kind`, but that might mask other unsafe uses of `kind`. I could add a `fn shallow_clone_kind` that is suppressed instead... Either way, I believe a race is not actually unsafe here, since the `kind` check is only trusted in safe cases (when inline or static), and otherwise, an atomic load is used.

PR #218, opened 2018-06-27
Fix `cargo doc` error on nightly caused by broken link to footnote

kpp:


kpp:
@carllerche please create a new release with this patch to fix https://github.com/tokio-rs/tokio/issues/437

PR #221, opened 2018-07-30
Add LocalBytes and LocalBytesMut for non-thread safe version

agemocui:
Added LocalBytes and LocalBytesMut to be used in shared-nothing case. The implementation is exactly the same as Bytes/BytesMut except that `usize` instead of `AtomicUsize` is used for `ref_count`.

See performance improved on 2 bench cases `clone_arc` and `split_off_and_drop` (2 more split_off calls are added into this bench).

Below is the result of benchmarks run on my laptop.
```text
Bytes
test clone_arc                      ... bench:      19,290 ns/iter (+/- 1,592)
test split_off_and_drop             ... bench:     125,821 ns/iter (+/- 11,190)

LocalBytes
test clone_arc                      ... bench:      11,903 ns/iter (+/- 1,347)
test split_off_and_drop             ... bench:      96,082 ns/iter (+/- 10,590)
```

### DIFF
--- a/benches/local_bytes.rs
+++ b/benches/local_bytes.rs
@@ -3,14 +3,14 @@
 extern crate bytes;
 extern crate test;
 
+use bytes::{BufMut, LocalBytes, LocalBytesMut};
 use test::Bencher;
-use bytes::{Bytes, BytesMut, BufMut};
 
 #[bench]
 fn alloc_small(b: &mut Bencher) {
     b.iter(|| {
         for _ in 0..1024 {
-            test::black_box(BytesMut::with_capacity(12));
+            test::black_box(LocalBytesMut::with_capacity(12));
         }
     })
 }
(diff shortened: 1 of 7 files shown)


dekellum:
Any way to make the Atom v usize a new generic parameter of the existing Bytes?  Or for compatibilty, new `GenBytes<T>` with `Bytes` becomes `pub type Bytes = GenBytes<AtomicUsize>`?

carllerche:
Thanks for doing this work.

At this current time, adding a current thread only version of `Bytes` is out of scope. That said, this could easily be provide in a crate maintained by you :). `bytes` provides `Buf` and `BufMut` traits, which provide the abstractions necessary to write code that can support both `Bytes` and `LocalBytes` from your crate.

Let me know if you have any further questions.

dekellum:
Question @carllerche : If the best way to do this is via generics (my prior suggestions), then if not in *bytes* crate (e.g. as you say "out of scope") then it becomes a fork, doesn't it?  Is forking a lesser evil in your view?

carllerche:
The Bytes type is not intended to handle all scenarios. It is intended to cover most pretty well and be easy. Adding a generic would go against this design goal. 

The intent has always been to expand on collection types in other crates. For example, I would like to see a rope implementation.

dekellum:
OK, thanks for clarifying your position.

agemocui:
@dekellum Actually I thought about the solution you mentioned. The reason I didn't go with that solution is that didn't feel good to expose GenBytes<T> and the RefCcount trait, especially not to expose constructor methods in GenBytes. 

@carllerche Understand your concerns. However, just feel like weird/uncomfortable to create a separate crate with 99% code copied from bytes, though there are not any license issues.

PR #229, opened 2018-09-30
Implement Buf for std::io::Take<&[u8]>.

smarnach:
I stumbled upon https://github.com/rust-lang/rust/issues/27269 and noticed that it is fixed now, so I added an implementation.  The implementation only covers `Take<&[u8]>`, but I can't think of any more general type this could be implemented for.


smarnach:
The build fails for Rust 1.15, which does not yet have the new methods.  So if this version of Rust still needs to be supported, we need to protect this by a feature flag, which probably isn't worth it.

carllerche:
I will track this for the next breaking change 👍 Thanks.

PR #231, opened 2018-11-15
Bring more attention to short reads/slices on Buff/BuffMut

vorner:
The property the Buff and BuffMut can return shorter slice is quite an
important detail. Nevertheless, while it is mentioned in the
documentation, the wording makes it relatively easy to overlook. This
tries to bring more attention to it.


vorner:
Hmm… changing only the doc comments should not trigger new address sanitizer errors I guess :-|

carllerche:
@vorner Thanks for the change!

The tsan warning is due to a change in std. I have a PR w/ a new whitelist.

PR #233, opened 2018-11-17
Use raw pointers for potentially racy loads

RalfJung:
Shared references assert immutability, so any concurrent write would be UB disregarding data race concerns. Better use raw pointers.


carllerche:
Thanks. Waiting for #234 to deal with the tsan error.

PR #236, opened 2018-11-20
Be clear about Inner::kind being deliberate UB

RalfJung:
Based on discussion in https://internals.rust-lang.org/t/bit-wise-reasoning-for-atomic-accesses/8853


carllerche:
It was merged into v0.4.x. master is temporarily on hold.

PR #239, opened 2018-12-18
Be clear about Inner::kind being deliberate UB

RalfJung:
Somehow when https://github.com/carllerche/bytes/pull/236 was merged, part of this change got lost?


RalfJung:
Ah no, you just ended up not pushing to the main branch what you pushed to the PR. Never mind.

PR #247, opened 2019-02-24
Use constants in bytes.rs test code

luaneyed:
I think using constants in test code is better for maintenance.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -2586,22 +2586,33 @@ fn original_capacity_from_repr(repr: usize) -> usize {
 
 #[test]
 fn test_original_capacity_to_repr() {
-    for i in 0..32 {
-      let cap = 1 << i;
+    let max_width = 32;
 
-      let expected = if i < MIN_ORIGINAL_CAPACITY_WIDTH {
-        0
-      } else if i < MAX_ORIGINAL_CAPACITY_WIDTH {
-        i + 1 - MIN_ORIGINAL_CAPACITY_WIDTH
-      } else {
-        MAX_ORIGINAL_CAPACITY_WIDTH - MIN_ORIGINAL_CAPACITY_WIDTH
-      };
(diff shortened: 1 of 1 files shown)


src/bytes.rs
```diff
@@ -2586,35 +2586,38 @@ fn original_capacity_from_repr(repr: usize) -> usize {
...
 #[test]
 fn test_original_capacity_to_repr() {
-    for &cap in &[0, 1, 16, 1000] {
-        assert_eq!(0, original_capacity_to_repr(cap));
-    }
-
-    for &cap in &[1024, 1025, 1100, 2000, 2047] {
-        assert_eq!(1, original_capacity_to_repr(cap));
-    }
+    for i in 0..32 {
```
carllerche: Thanks for the PR. I see where are coming from.

If I am reading it right, this change seems to restrict the numbers being tested to powers of two and powers of 2 + 1. Before, there were more types of numbers being tested.

What are your thoughts?

luaneyed: You read it right.

I think there is no right answer on deciding test coverage of the function.
Would it be enough if I add 0, 1000, 1100, 2000, 2047?
Or we can consider testing some random-generated numbers.

carllerche: You are right... there is no right answer :) There also is the risk of "reimplementing" the original algorithm and introducing bugs in the test.

If you want to pursue the PR, I would merge it w/ the added numbers you listed.

luaneyed: Please check the changed test code!
I tried to use `MIN_ORIGINAL_CAPACITY_WIDTH` for the added numbers, too.

carllerche: 👍

carllerche:
Got a failing test!

luaneyed:
Oh, it's my first cross-compiling issue 😂
I fixed it!

Issue #248, opened 2019-03-02
Buf (and BufMut) for VecDeque


vorner:
Hello

I'd like to have an idea validated before I jump in and implement it.

Let's say I have a decoding routine for a record and it requires `IntoBuf`. I don't know how large the record is going to be ahead of decoding it, but either there's an upper limit or it's able to tell me „not enough, give me more“. I have a huge file of these records (specifically, these are length-prefixed protobuf encoded messages and I want to use prost to decode it).

I could just slurp the whole file into memory and then repeatedly try to decode the message, like this:

```rust
let mut buffer = Vec::new();
file.read_to_end(&mut buffer)?;
let mut buffer = Cursor::new(buffer);
while buffer.position() < buffer.len() as usize {
  let _msg = Message::decode_length_delimited(&mut buffer)?;
}
```

This, however, has the downside of putting the whole huge file into RAM.

I'd like something more in the lines of:

```rust
let mut buffer = SomeBuffer::new();
while !buffer.is_empty() {
    // If there's less than LOW_MARK, tries to read enough to reach HIGH_MARK.
    // It's outside of the `Buf` trait methods, because this can return errors.
    // Stores an eof mark if it reaches end of the file and is_empty will return true only
    // if it contains no bytes and eof was reached.
    buffer.refill(HIGH_MARK, LOW_MARK, &mut file)?;
    let _msg = Message::decode_length_delimited(&mut buffer)?;
}
```

Now, this is somewhat higher level API, but I was wondering how to implement it. My first thought would be `BytesMut` should be able to do it (even after reading code related to tokio::codec and tokio::io suggest it should be the way forward). However, I don't see one thing. The `BytesMut` tries to have everything in a continuous chunk of memory. So, if I have the bad luck of never consuming the full buffer and leaving something in there every time, an unused space on the left will gradually grow while I'll keep allocating on the right. Or how does this work? If it does work by some trick I don't see, is it possible to extend the documentation with an example and explanation?

However, then I discovered that what I need is mostly a FIFO for bytes and that's what VecDeque does. I believe implementing `Buf` for it should be easy. Regarding this, do you see any problem with having VecDeque implement it? Does it make sense? If so, I'll send a PR.

carllerche:
Yep. I think that you are on the right path. The crate used to contain a RingBuf type that had a fixed capacity. I think your best bet would be to define a type and wrap VecDeque.

[...]

PR #249, opened 2019-03-04
Implementation of Buf for VecDeque

vorner:
Since it can be easily re-filled.

I'll think about how (if possible) to implement BufMut too.

Related to #248.

### DIFF
--- a/src/buf/mod.rs
+++ b/src/buf/mod.rs
@@ -24,6 +24,7 @@ mod into_buf;
 mod iter;
 mod reader;
 mod take;
+mod vec_deque;
 mod writer;
 
 pub use self::buf::Buf;
--- a/src/buf/vec_deque.rs
+++ b/src/buf/vec_deque.rs
@@ -1,0 +1,39 @@
+use std::collections::VecDeque;
+
+use super::Buf;
+
+impl Buf for VecDeque<u8> {
+    fn remaining(&self) -> usize {
+        self.len()
+    }
+
+    fn bytes(&self) -> &[u8] {
+        let (s1, s2) = self.as_slices();
+        if s1.is_empty() {
+            s2
+        } else {
+            s1
+        }
+    }
+
+    fn advance(&mut self, cnt: usize) {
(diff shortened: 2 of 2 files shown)


vorner:
Ah, I haven't read the comment on #248 yet… ok.

carllerche:
The primary reason I am not a huge fan of implementing `Buf` and `BufMut` for `VecDeque<u8>` is that it feels weird to me that `bytes()` and `bytes_mut()` return different ranges. This is also why `RingBuf` used to have `reader()` and `writer()` accessors.

carllerche:
BTW, if you release a crate, I'd be happy to include a section in the README that links to additional implementations of Buf / BufMut.

vorner:
Isn't the different ranges somewhat a problem of `Vec<u8>` too? Well, the `IntoBuf` wraps it into a `Cursor`, but still.

Hmm, it seems it isn't reasonably possible to implement `BytesMut` for VecDeque anyway ‒ it doesn't expose a way to access the unused part of allocation or raw pointers inside it.

I'll keep the offer with readme in mind, but considering the above, I'll have to think if I even want to create the crate. It might still work in the use case I described, though, even without that ability.

carllerche:
You are probably right about the inconsistency. I guess the question should be if it is reasonable for single types to impl both Buf and BufMut such that data is read from the front and written to the back.

carllerche:
Cc @seanmonstar

carllerche:
@vorner I'll re-open. I think you are probably right.

carllerche:
@vorner Ok, this looks good. Do you want to add a `BufMut` impl or do you want it merged as is?

vorner:
Probably as it is. As I said, I don't see a way to implement `BufMut` right now. If I come up with one later on, I'll open another PR.

Anyway, if this is becomes a problem or too much work when going to the 0.5, it's always possible to drop the support or you can ask me to port it. I'm not really a good maintainer of stuff, but I should at least be able to take care of what I wrote.

PR #255, opened 2019-04-14
no_std support (with default-enabled "std" cargo feature)

tarcieri:
This PR revives #135 now that the [`alloc` crate is stable](https://github.com/rust-lang/rust/pull/59675), which unlocks using `bytes` in embedded RTOS environments or other embedded environments with enough RAM to warrant a heap (e.g. on SAMD51 uCs that have 192kB RAM).

Like the original #135 actual `alloc` support will be added in a subsequent PR, and this is a first pass to add initial `no_std` support and sanity check the approach.

In absence of upstream support for this, I've had to create [forks of bytes](https://github.com/microcrates/bytes) and [forks of downstream consumers of bytes](https://github.com/microcrates/prost), so it'd be really swell to get this all upstream rather than forking!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ matrix:
     - env: RUST_TEST_THREADS=1 TARGET=powerpc-unknown-linux-gnu
     - env: RUST_TEST_THREADS=1 TARGET=powerpc64-unknown-linux-gnu
 
+    # Ensure crate compiles without default features (i.e. for no_std)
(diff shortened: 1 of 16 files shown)


tarcieri:
Note: I'd also be happy to pursue this on the 0.5.x branch only, as I think there are some opportunities there to address other tough problems like the pervasive use of `io::Cursor`

tarcieri:
Obsoleted by #281

PR #256, opened 2019-04-14
alloc support (for no_std environments)

tarcieri:
This PR revives #153 and includes #255. I'm reopening it because the [`alloc` crate is now stable](https://github.com/rust-lang/rust/pull/59675), which should hopefully address the previous concerns about #153 being `nightly`-dependent.

It gates all allocator-dependent features on the `alloc` cargo feature, enabling use of much of the functionality in `no_std` environments which have a heap and defined allocator, but not full support for 'std'.

Unlike previous attempts, this PR runs the parts of the test suite which do not require `std` as part of CI, ensuring that `alloc` support is properly tested. For now this requires `nightly`, but when `alloc` stabilization lands in the `stable` channel, it should be possible to always test `alloc` support in all environments.

### DIFF
--- a/src/buf/into_buf.rs
+++ b/src/buf/into_buf.rs
@@ -1,5 +1,8 @@
 use super::{Buf};
+#[cfg(feature = "std")]
+use prelude::*;
 
+#[cfg(feature = "std")]
 use std::io;
 
 /// Conversion into a `Buf`
@@ -55,6 +58,7 @@ impl<T: Buf> IntoBuf for T {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a [u8] {
     type Buf = io::Cursor<&'a [u8]>;
 
@@ -63,6 +67,7 @@ impl<'a> IntoBuf for &'a [u8] {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a mut [u8] {
     type Buf = io::Cursor<&'a mut [u8]>;
 
@@ -71,6 +76,7 @@ impl<'a> IntoBuf for &'a mut [u8] {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a str {
     type Buf = io::Cursor<&'a [u8]>;
 
@@ -79,6 +85,7 @@ impl<'a> IntoBuf for &'a str {
     }
 }
 
+#[cfg(feature = "std")]
 impl IntoBuf for Vec<u8> {
     type Buf = io::Cursor<Vec<u8>>;
 
@@ -87,6 +94,7 @@ impl IntoBuf for Vec<u8> {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a Vec<u8> {
     type Buf = io::Cursor<&'a [u8]>;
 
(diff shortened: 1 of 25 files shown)


src/buf/into_buf.rs
```diff
@@ -63,6 +67,7 @@ impl<'a> IntoBuf for &'a [u8] {
     }
 }
 
+#[cfg(feature = "std")]
 impl<'a> IntoBuf for &'a mut [u8] {
     type Buf = io::Cursor<&'a mut [u8]>;
```
tarcieri: To really support `alloc` properly I think these `impl`s (both this one and the one below for `&'a str` need to change to not use an `io::Cursor`), however as `Buf` is an associated type, I imagine it's considered part of the public API and changes to the underlying `Buf` type would be considered breaking.

As a breaking change I think ideally we could fix this by eliminating all usages of `io::Cursor` from the crate (which, I believe, was @carllerche's plan for v0.5.x anyway). However I'd like to stress that even without these impls, bytes is still useful, because downstream `no_std` + `alloc` consumers of `bytes` can define their own buffer type (possibly just vendoring `io::Cursor` from `libstd` and using the existing `std`-dependent impls) and then `imp FromBuf` and `impl IntoBuf` for their own type. This would be my planned approach for now if this PR were to be merged.

src/lib.rs
```diff
@@ -70,9 +70,18 @@
 
 #![deny(warnings, missing_docs, missing_debug_implementations)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.4.12")]
+#![no_std]
 
+// TODO: remove this when alloc stabilization lands in the 'nightly' channel
+#![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(alloc))]
```
tarcieri: Looks like `alloc` stabilization isn't quite in the latest nightly on Travis CI, but as soon as it is this can be removed.

src/prelude.rs
```diff
@@ -0,0 +1,12 @@
...
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+pub use alloc::boxed::Box;
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+pub use alloc::string::String;
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+pub use alloc::vec::Vec;
+
+#[cfg(feature = "std")]
+pub use std::prelude::v1::*;
```
tarcieri: This looks a little bit weird because we pull these types in from different places depending on the precise set of features enabled, so it doesn't look purely additive, however `std` just re-exports the `alloc` types as a facade, so it winds up being purely additive.

src/lib.rs
tarcieri: Update: removed now that this has been stabilized in the nightly channel

carllerche:
Thanks. At this point, focus is shifting towards the next breaking release. I’ll aim to get this included.

mzabaluev:
Please take a look at #269, my design proposal to abstract away the allocator. `no_std` applications may want to use completely custom allocators, not necessarily backed by a global allocator.

tarcieri:
@mzabaluev Looks good to me! I'd support that approach over what's in this PR

tarcieri:
Obsoleted by #281

PR #257, opened 2019-06-05
Deprecate BytesMut::take(), renamed to split()

seanmonstar:
In 0.5, we wish to implement `Buf` for `BytesMut`, and there is a name conflict with `Buf::take()`. To prepare for it now, rename the current `BytesMut::take()` to `BytesMut::split()`. The docs help with the name choice: it's the same as `bytes.split_to(bytes.len())`.


carllerche:
Why worry about deprecation, etc... at this point? we can make breaking changes on master and move forward.

PR #263, opened 2019-06-07
Replace iovec with std::io::IoSlice

### LINKED ISSUE #224: Replace iovec with std::io::IoSlice
This would allow avoiding the additional dependencies.

seanmonstar:
Closes #224

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,8 @@ dist: trusty
 language: rust
 services: docker
 sudo: required
-rust: stable
+#rust: stable
+rust: beta # we need 1.36, which is still beta
 
 env:
   global:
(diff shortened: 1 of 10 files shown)


archseer:
Should these be renamed to match the iovec methods in the standard library? `_vec` -> `_vectored`
`stdlib` also returns `Result<usize>` instead of `usize`, but I'm not sure that's applicable here.

seanmonstar:
> `_vec` -> `_vectored`

I think this makes sense, it's clearer they are meant for "vectored IO", not for `Vec`.

Thoughts @carllerche?

carllerche:
sure.

PR #273, opened 2019-07-24
Update travis config to test on stable

taiki-e:


taiki-e:
@carllerche I'd like to try that. azure seems faster than travis :)

carllerche:
I added azure to the project: https://github.com/tokio-rs/bytes/tree/azure-pipelines the yaml config just needs to be setup.

carllerche:
Azure page: https://dev.azure.com/tokio-rs/bytes/_build?definitionId=3&_a=summary

PR #274, opened 2019-07-25
Update Bytes to Rust 2018

taiki-e:
Now that minimum supported Rust version has become 1.36, we can do this.


carllerche:
(you should have permission to merge this PR too... please let me know if you don't).

PR #275, opened 2019-07-25
Set up CI with Azure Pipelines

carllerche:


taiki-e:
Hmm... cross doesn't seem to support Azure Pipelines yet: https://github.com/rust-embedded/cross/pull/169

taiki-e:
There seems to be [a workaround](https://nbsoftsolutions.com/blog/azure-pipelines-for-rust-projects#workaround-cross-compilation). I'm not sure if we should use it in the end, but I'll try it.

ci/azure-cross-compile.yml
```diff
@@ -0,0 +1,52 @@
...
+    matrix:
+      i686:
+        vmImage: ubuntu-16.04
+        target: i686-unknown-linux-gnu
+      armv7:
+        vmImage: ubuntu-16.04
+        target: armv7-unknown-linux-gnueabihf
+      powerpc:
+        vmImage: ubuntu-16.04
+        target: powerpc-unknown-linux-gnu
```
taiki-e: I removed `RUST_TEST_THREADS=1` because it looks like cross is already setting it.
* https://github.com/rust-embedded/cross/blob/540dc4f5a1ca8d6d95bbf4da66ce763a2352d36f/docker/powerpc-unknown-linux-gnu/Dockerfile#L53
* https://github.com/rust-embedded/cross/blob/540dc4f5a1ca8d6d95bbf4da66ce763a2352d36f/docker/powerpc64-unknown-linux-gnu/Dockerfile#L49

taiki-e:
@carllerche This PR should be ready for review.

The build time seems to have been reduced from 5-6 to 3-4 minutes.

Although I did not actually test deployment, I think that it works well because `ci/azure-deploy-docs.yml` and `ci/azure-install-rust.yml` are basically the same as mio's (The only difference is that {Mio,mio} has been replaced by {Bytes,bytes}.).

Also, I removed `ci/before_deploy.sh` because [they did not seem to be actually used](https://travis-ci.org/tokio-rs/bytes/jobs/502766885).

Cargo.toml
```diff
@@ -16,10 +16,10 @@ readme        = "README.md"
 keywords      = ["buffers", "zero-copy", "io"]
 exclude       = [
     ".gitignore",
-    ".travis.yml",
-    "deploy.sh",
+    "azure-pipelines.yml",
```
carllerche: Unrelated, but I wonder if the `exclude` is still relevant with modern Cargo. At the very least we probably should include tests?

taiki-e: IIRC, (at least in modern cargo) the purpose of the `exclude` field is to prevent the size of the crate from being bloated by large files like images. So I think we can remove this. (By the way, By the way, as we do not currently have a `test` directory or `bench` directory, the current `exclude` list does little work. We have `tests` directory and `benches` directory.)

carllerche: Yeah, the exclude statement is really really old...

carllerche:
@taiki-e Funny enough, I can't +1 this because I opened the PR... So, I am +1 and will force merge this :)

carllerche:
Thanks again!

taiki-e:
Hmm... seems to deploy docs was failed: https://dev.azure.com/tokio-rs/bytes/_build/results?buildId=1699&view=logs&jobId=bbfa2fa6-a202-51ab-3559-6013e9e5f686
```text
remote: Invalid username or password.
fatal: Authentication failed for 'https://github.com/tokio-rs/bytes/'
```

carllerche:
Ah.. I will have to fix this.

PR #276, opened 2019-07-27
Remove i128 feature

taiki-e:
byteorder v1.3 (or later) automatically detects the rustc version and enables i128 support.

Refs: https://github.com/BurntSushi/byteorder/pull/139/commits/168648f5c2339a368a3273af3a444fca0d6e1be1


carllerche:
Waiting to see if the deploy docs task is fixed [here](https://dev.azure.com/tokio-rs/bytes/_build/results?buildId=1701).

carllerche:
Trying again: https://dev.azure.com/tokio-rs/bytes/_build/results?buildId=1703

taiki-e:
Thanks for the fix.

PR #278, opened 2019-07-30
impl `LowerHex`/`UpperHex` for `Bytes`/`BytesMut`

nanpuyue:


src/hex.rs
```diff
@@ -0,0 +1,37 @@
+use crate::{Bytes, BytesMut};
+use std::fmt::{Formatter, LowerHex, Result, UpperHex};
+
+struct BytesRef<'a>(&'a [u8]);
+
+impl<'a> LowerHex for BytesRef<'a> {
```
DoumanAsh: Since it anyway needs macros to impl for multiple types at once, wouldn't be easier to just define impl without `BytesRef`?

src/bytes.rs
```diff
@@ -1666,7 +1666,7 @@ impl<'a> From<&'a [u8]> for BytesMut {
             BytesMut::new()
         } else if len <= INLINE_CAP {
             unsafe {
-                let mut inner: Inner = mem::uninitialized();
+                let mut inner: Inner = mem::MaybeUninit::uninit().assume_init();
```
quininer: This is not the correct usage of `MaybeUninit`.

see https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#examples

DoumanAsh: it is fine though, since it is equivalent of original code

quininer: This is not fine, otherwise `uninitialized` will not be deprecated.

DoumanAsh: it is deprecated because it was easy to misuse and some edge cases, there it is used just fine

src/hex.rs
nanpuyue: In this case, we can merge `BsDebug` and `BytesRef` later, and avoid code bloat.

nanpuyue: Let me know if my understanding is wrong.

DoumanAsh: I suspect in the end it would be inlined, but now I think about it, it is a little difference

DoumanAsh:
Thanks

PR #280, opened 2019-08-10
Remove byteorder dependency

### LINKED ISSUE #279: Remove byteorder-dependency?
The `byteorder`-crate is not strictly needed since 1.32 and could be removed. Dropping a dependency is worth the change IMHO. Is there interest in doing so?

DoumanAsh:
This what it takes to remove the dependency.
It is a bit difficult to make it generic to retrieve le/be from array because std methods only allow plain sized array.
But it shouldn't be performance hit because byteorder copies slices always

Fixes #279

### DIFF
--- a/src/buf/buf.rs
+++ b/src/buf/buf.rs
@@ -3,23 +3,24 @@ use super::{IntoBuf, Take, Reader, FromBuf, Chain};
 use std::{cmp, io::IoSlice, ptr, mem};
 
 macro_rules! buf_get_impl {
-    ($this:ident, $size:expr, $conv:path) => ({
+    ($this:ident, $typ:tt::$conv:tt) => ({
+        const SIZE: usize = mem::size_of::<$typ>();
          // try to convert directly from the bytes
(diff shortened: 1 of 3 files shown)


src/buf/buf.rs
```diff
@@ -715,7 +720,7 @@ pub trait Buf {
     ///
     /// This function panics if there is not enough remaining data in `self`.
     fn get_f32(&mut self) -> f32 {
-        buf_get_impl!(self, 4, BigEndian::read_f32);
+        unsafe { *(&Self::get_u32(self) as *const u32 as *const f32) }
```
taiki-e: Perhaps you can get rid of these unsafe block.

https://github.com/rust-lang/rust/issues/60446
> unlike corresponding integer methods, these can be implemented without unsafe code. So the need for having them in the standard library feels less pressing. At the same time, it might not be obvious to everyone that to_bits and from_bits can be combined with integer methods in this way.

DoumanAsh: Yep, you're right

Issue #269, opened 2019-06-25
Support for custom allocators/ref counting


mzabaluev:
Allocation and reference counting of the buffers could be abstracted behind a generic type parameter. This can enable:

- Lightweight non-atomic reference counting for single-threaded use (raised previously as #200).
- Fast task-bound, non-global allocators.
- Use in `no_std` configurations with a custom allocator.

Here's a sketch of a possible abstraction API:

```rust
use core::alloc::AllocErr;
use core::ptr::NonNull;
#[cfg(feature = "system_alloc")]
use std::alloc::System as SystemAlloc;

/// Represents results of an allocation: a handle that represents
/// the allocated block and is used to keep track of
/// the references, and the buffer's total capacity (which may be larger
/// than requested).
pub struct BufAllocation<H>(pub H, pub usize);

/// Handle to a shared buffer descriptor.
///
/// A type implementing this trait is an opaque handle that provides
/// low-level access to a shared data buffer and is used by higher-level
/// container implementations to keep track of references to the buffer.
/// The implementing type should also implement `Clone` by producing another
/// handle referring to the same buffer.
///
/// Neither the handle, nor a shared descriptor structure that it possibly
/// points to, own buffer's data. If the handle is dropped without the
/// `release` method called on it first, the implementation will leak the data,
/// the reference, or both.
///
/// The implementation type can exhibit specific thread safety
/// characteristics with regard to `Send` and `Sync`.
/// For global thread-safe allocators, it can be modeled as an `Arc`
/// pointing to a structure with the data pointer and the buffer's
/// allocated size.
///
/// It is up to the implementation whether to allocate the descriptor
/// and the data block in a single contiguous allocation or separately.
/// For arena-like allocators that never free individual allocations,
/// `BufHandle` could contain just the data pointer and a pointer to the
/// shared arena state.
///
pub trait BufHandle: Sized {
    fn ptr(&self) -> NonNull<u8>;

    /// Release the reference on the buffer represented by `self`
    /// and deallocate buffer data if the reference was unique.
    /// The typical place for calling this method is the `Drop`
    /// implementation of a type containing the handle referenced by `self`.
    ///
    /// # Safety
    ///
    /// The only safe way to use this function is to call it exactly once
    /// per the lifetime of the handle, without calling any other methods
    /// afterwards before the handle is dropped. The caller also has to
    /// ensure that no use of the allocated buffer data occurs after the
    /// last reference to it is released.
    ///
    unsafe fn release(&mut self);

[...]

PR #281, opened 2019-08-13
Add no_std support, by adding an `std` feature

cbeck88:
I was really hopeful initially that this could be done without
having an `std` feature, but it turns out that `bytes` has a bunch
of special additional integration with `std::io` stuff, e.g.
`std::io::Reader` and `std::io::IoSlice`.

To make the library work as `no_std` we add an `std` feature which
is on by default. When it is off, we compile as `no_std` and make
parts of the API that require `std::io` conditional on the `std`
feature.

### DIFF
--- a/src/buf/vec_deque.rs
+++ b/src/buf/vec_deque.rs
@@ -1,4 +1,7 @@
+#[cfg(not(feature = "std"))]
 use alloc::collections::VecDeque;
+#[cfg(feature = "std")]
+use std::collections::VecDeque;
 
 use super::Buf;
 
--- a/src/buf/vec_deque.rs
+++ b/src/buf/vec_deque.rs
@@ -1,7 +1,4 @@
-#[cfg(not(feature = "std"))]
 use alloc::collections::VecDeque;
-#[cfg(feature = "std")]
-use std::collections::VecDeque;
 
 use super::Buf;
 
--- a/src/buf/vec_deque.rs
+++ b/src/buf/vec_deque.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use alloc::collections::VecDeque;
 
 use super::Buf;
 
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -11,7 +11,7 @@ use core::iter::{FromIterator, Iterator};
 #[cfg(not(feature = "std"))]
 use alloc::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
 #[cfg(feature = "std")]
-use std::borrow::{Borrow, BorrowMut};
+use std::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
 
 /// A reference counted contiguous slice of memory.
 ///
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -8,10 +8,7 @@ use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
 use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
 use core::iter::{FromIterator, Iterator};
 
-#[cfg(not(feature = "std"))]
 use alloc::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
-#[cfg(feature = "std")]
-use std::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
 
 /// A reference counted contiguous slice of memory.
 ///
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -3,11 +3,15 @@ use crate::buf::IntoIter;
 use crate::debug;
 
 use core::{cmp, fmt, mem, hash, slice, ptr, usize};
-use alloc::borrow::{Borrow, BorrowMut};
 use core::ops::{Deref, DerefMut, RangeBounds};
 use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
 use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
 use core::iter::{FromIterator, Iterator};
+
+#[cfg(not(feature = "std"))]
+use alloc::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
+#[cfg(feature = "std")]
+use std::borrow::{Borrow, BorrowMut};
 
 /// A reference counted contiguous slice of memory.
 ///
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -2,12 +2,12 @@ use crate::{Buf, BufMut};
 use crate::buf::IntoIter;
 use crate::debug;
 
-use std::{cmp, fmt, mem, hash, slice, ptr, usize};
-use std::borrow::{Borrow, BorrowMut};
-use std::ops::{Deref, DerefMut, RangeBounds};
-use std::sync::atomic::{self, AtomicUsize, AtomicPtr};
-use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
-use std::iter::{FromIterator, Iterator};
+use core::{cmp, fmt, mem, hash, slice, ptr, usize};
+use alloc::borrow::{Borrow, BorrowMut};
+use core::ops::{Deref, DerefMut, RangeBounds};
+use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
+use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
+use core::iter::{FromIterator, Iterator};
 
 /// A reference counted contiguous slice of memory.
 ///
@@ -316,10 +316,10 @@ struct Inner {
 }
 
 // Thread-safe reference-counted container for the shared storage. This mostly
-// the same as `std::sync::Arc` but without the weak counter. The ref counting
+// the same as `core::sync::Arc` but without the weak counter. The ref counting
 // fns are based on the ones found in `std`.
 //
-// The main reason to use `Shared` instead of `std::sync::Arc` is that it ends
+// The main reason to use `Shared` instead of `core::sync::Arc` is that it ends
 // up making the overall code simpler and easier to reason about. This is due to
 // some of the logic around setting `Inner::arc` and other ways the `arc` field
 // is used. Using `Arc` ended up requiring a number of funky transmutes and
@@ -527,7 +527,7 @@ impl Bytes {
     /// Requires that `begin <= end` and `end <= self.len()`, otherwise slicing
     /// will panic.
     pub fn slice(&self, range: impl RangeBounds<usize>) -> Bytes {
-        use std::ops::Bound;
+        use core::ops::Bound;
 
         let len = self.len();
 
@@ -857,7 +857,7 @@ impl Bytes {
     /// assert_eq!(iter.next().map(|b| *b), Some(b'c'));
     /// assert_eq!(iter.next(), None);
     /// ```
-    pub fn iter<'a>(&'a self) -> std::slice::Iter<'a, u8> {
+    pub fn iter<'a>(&'a self) -> core::slice::Iter<'a, u8> {
         self.bytes().iter()
     }
 }
@@ -1031,7 +1031,7 @@ impl IntoIterator for Bytes {
 
 impl<'a> IntoIterator for &'a Bytes {
     type Item = &'a u8;
-    type IntoIter = std::slice::Iter<'a, u8>;
+    type IntoIter = core::slice::Iter<'a, u8>;
 
     fn into_iter(self) -> Self::IntoIter {
         self.as_ref().into_iter()
@@ -1539,7 +1539,7 @@ impl BytesMut {
     /// assert_eq!(iter.next().map(|b| *b), Some(b'c'));
     /// assert_eq!(iter.next(), None);
     /// ```
-    pub fn iter<'a>(&'a self) -> std::slice::Iter<'a, u8> {
+    pub fn iter<'a>(&'a self) -> core::slice::Iter<'a, u8> {
         self.bytes().iter()
     }
 }
@@ -1780,7 +1780,7 @@ impl IntoIterator for BytesMut {
 
 impl<'a> IntoIterator for &'a BytesMut {
     type Item = &'a u8;
-    type IntoIter = std::slice::Iter<'a, u8>;
+    type IntoIter = core::slice::Iter<'a, u8>;
 
     fn into_iter(self) -> Self::IntoIter {
         self.as_ref().into_iter()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -73,6 +73,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+#[cfg(not(feature = "std"))]
 extern crate alloc;
 
 pub mod buf;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -71,6 +71,10 @@
 #![deny(warnings, missing_docs, missing_debug_implementations, rust_2018_idioms)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.5.0")]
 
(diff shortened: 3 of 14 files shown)


cbeck88:
Hi -- I'm interested in getting `bytes` to work and build as a ` no_std` library relying on the newly-stabilized `alloc` crate to get `Vec` etc.

My motivation for doing this is that we want to use `prost` for serialization inside of an SGX enclave. Inside SGX you may have an allocator but you can't talk to the OS directly. `bytes` is a dependency of `prost`.

I think it's pretty unlikely that all of `tokio` could be ported inside of sgx, but if you are interested in `no_std` support in some of these low-hanging dependencies like bytes that are really fundamental, it will make them maximally portable and help the ecosystem.

LMK if you are interested in a patch like this! Thanks

tarcieri:
@cbeck88 you should probably check out:

#269
#256 
#255 
#153
#135

cbeck88:
@tarcieri My PR is very simple -- `alloc` has been standardized, and master is already on edition 2018. So the only thing we have to do is get things from the right namespace, and feature-gate the use of `std`-only trait bounds. This is a simple and mechanical change. This is not new functionality or code change at all, so it's hopefully not controversial.

#269 seems orthogonal and way beyond my goals
#256 and #255 are both stale since iovec was dropped. They also do some complex stuff with

```
#[cfg(feature = "std")]
use prelude::*;
```

that seems totally unnecessary.

#153 and #135 are both from 2 years ago. Is there anything from these threads in particular that you want to draw attention to? They look moot to me.

tarcieri:
I pointed it out to note that there are other open PRs for this same general issue, for the ample historical discussion if nothing else. #269 describes solutions which can work for heapless `no_std` users (which this PR doesn’t help)

re: a crate-local prelude

> that seems totally unnecessary

It’s necessary for compatibility with Rust versions prior to 1.35. It’s certainly much cleaner to support 1.35+, and I’m not sure what MSRV plans are for bytes 0.5.

cbeck88:
@tarcieri I see -- I had mistakenly thought that `edition: 2018` already put us past 1.35
Yeah if MSRV for bytes 0.5 is 1.35 or less then my patch needs more work

tarcieri:
Perhaps @carllerche can tell us what he's thinking re: `bytes` 0.5 MSRV?

taiki-e:
MSRV of Bytes 0.5 is 1.36.
https://github.com/tokio-rs/bytes/blob/234d814122d6445bdfb15f635290bfc4dd36c2eb/azure-pipelines.yml#L15-L23

DoumanAsh:
@cbeck88 Could you please resolve current conflicts?

Master branch targets 2018 so it is perfectly fine as PR.

I'm still not sure what sort of no_std support we should have, but I believe support for global allocator would be good first step

src/lib.rs
```diff
@@ -71,6 +71,10 @@
 #![deny(warnings, missing_docs, missing_debug_implementations, rust_2018_idioms)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.5.0")]
 
+#![cfg_attr(not(feature = "std"), no_std)]
```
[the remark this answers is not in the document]
DoumanAsh: `core` is always imported and `std` should be used only with `std` flag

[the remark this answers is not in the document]
DoumanAsh: `#![cfg_attr(not(feature = "std"), no_std)]` is simple enough.
I do not think it warrants unnecessary extern crate

src/bytes.rs
```diff
@@ -2,12 +2,12 @@ use crate::{Buf, BufMut, IntoBuf};
...
 use crate::debug;
 
-use std::{cmp, fmt, mem, hash, slice, ptr, usize};
-use std::borrow::{Borrow, BorrowMut};
-use std::ops::{Deref, DerefMut, RangeBounds};
-use std::sync::atomic::{self, AtomicUsize, AtomicPtr};
-use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
-use std::iter::{FromIterator, Iterator};
+use core::{cmp, fmt, mem, hash, slice, ptr, usize};
+use alloc::borrow::{Borrow, BorrowMut};
```
DoumanAsh: alloc is imported only on no_std target, do you know if it is always implicitly imported on std?
AFAIK in past I remember I needed to `use std as alloc` to make it work

DoumanAsh:
@cbeck88 I made some changes to your PR.
I believe we don't need to link against `alloc` when `std` is already present

cbeck88:
@DoumanAsh 

> I believe we don't need to link against alloc when std is already present

okay, but to my understanding there is no separate target to "link against". the symbols exported in std are simply re-exports of what is in `alloc`. so i think it's simpler to just always take the symbols from `alloc` and not do `if std use std::...`, because that just adds boilerplate that literally doesn't do anything.

that said, i'm happy with these changes if they make this PR acceptable to reviewers

src/bytes.rs
cbeck88: thanks, I guess the situation is more complex than I realized

DoumanAsh:
Hm maybe you're right, I usually do not link to alloc because it is a bit unclear about it's status when building in std enviornment

src/lib.rs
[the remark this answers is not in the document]
cbeck88: done

cbeck88:
@tarcieri @DoumanAsh let me know if you guys see anything else to improve here, i don't have plans to make further changes at this time

src/buf/vec_deque.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(not(feature = "std"))]
+use alloc::collections::VecDeque;
+#[cfg(feature = "std")]
 use std::collections::VecDeque;
```
tarcieri: `std` is just re-exporting the `VecDeque` type from `alloc`, so this sort of gating isn't really necessary

src/bytes.rs
```diff
@@ -2,12 +2,16 @@ use crate::{Buf, BufMut};
...
+use core::{cmp, fmt, mem, hash, slice, ptr, usize};
+use core::ops::{Deref, DerefMut, RangeBounds};
+use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
+use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
+use core::iter::{FromIterator, Iterator};
+
+#[cfg(not(feature = "std"))]
+use alloc::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
+#[cfg(feature = "std")]
+use std::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
```
tarcieri: Another place where types could be consistently imported via `alloc`

src/buf/vec_deque.rs
cbeck88: @tarcieri i agree with you, but the goal was to address

> I believe we don't need to link against alloc when std is already present

if in `std` mode we don't want to refer to the `alloc` crate at all, then this is the pattern we have to use, AFAIU

cbeck88: the lines you are commenting were added in @DoumanAsh commit: https://github.com/tokio-rs/bytes/pull/281/commits/2f3a5b0982bb0a29abe61942585d4bda6c5cb0a4

src/lib.rs
```diff
@@ -71,6 +71,14 @@
 #![deny(warnings, missing_docs, missing_debug_implementations, rust_2018_idioms)]
 #![doc(html_root_url = "https://docs.rs/bytes/0.5.0")]
 
+#![no_std]
+
+#[cfg(feature = "std")]
+extern crate std;
+
+#[cfg(not(feature = "std"))]
+extern crate alloc;
```
tarcieri: If this is targeting 1.36+, you can just use `extern crate alloc` and be done with it. Either way you're linking `liballoc`, it's just in one case you're going through the `std` facade.

Also note in #255 I had some of the core functionality of `bytes` available in a no-`alloc` context, which could be useful in conjunction with crates like [`heapless`](https://docs.rs/heapless/latest/heapless/)

src/buf/vec_deque.rs
cbeck88: lmk if you both think it's better to just use `no_std` and `extern crate alloc;` in all configurations, with `extern crate std` as well in std configuration, i can change it back to that way

tarcieri: What's the rationale for that?

cbeck88: i suspect that there is ultimately no difference either way

tarcieri: @cbeck88 that'd be my preference. Unless you're trying to support MSRV < 1.36 I don't see the point in doing otherwise

cbeck88: @DoumanAsh do you agree?

src/lib.rs
cbeck88: > Also note in #255 I had some of the core functionality of bytes available in a no-alloc context, which could be useful in conjunction with crates like heapless

I didn't understand this, doesn't bytes rely pretty fundamentally on `vec`? What parts of this crate were you going to build without pulling in `vec`? Just the `buf` and `buf_mut` parts?

What would you think about putting these trait objects in a smaller crate that would not have `alloc` dependency in any configuration? That's my instinct anyways

cbeck88: Hmm maybe that smaller crate would still need some flags I guess, on looking at `buf_impl.rs` again

cbeck88: @tarcieri yeah i guess i see what you are saying

do you want to see an `alloc` feature as well here?

tarcieri: The idea would be to use e.g. `heapless::Vec` in lieu of `alloc::vec::Vec. Probably a distraction to bring them up in this PR, sorry. Would be good to land this and circle back on full `no_std` support.

tarcieri: I'd say you don't need to add an `alloc` feature until there's a useful `no_std` mode without it, and that's probably best saved for a separate PR dealing with it specifically.

cbeck88: if we ultimately are going to want an alloc feature then it would be nice to figure that out beforehand because at this revision, `default-features = false` is the alloc configuration, so it would be a breaking change for anyone with that in their Cargo.toml if we add alloc feature later, and `default-features=false` becomes core-only

cbeck88: imo it might be nice to put the `Buf` traits in their own crate if core-only / heapless is a goal, that would reduce the scope of bytes crate to, providing the wrapper around `Vec` that implements those traits, and reduces the configuration surface area of `bytes` crate

tarcieri: If you want to tackle it now, I'm not going to complain 😉

A crate containing only the core traits sounds good to me

cbeck88: (sorry for confusion, my previous 4 comments were written without any of your comments being delivered from some reason)

i think i will play around with the smaller crate idea later, but i'm gonna hold off on it, and an alloc feature, for this PR

src/buf/vec_deque.rs
cbeck88: @DoumanAsh i think it's also worth noting that the RFC that stabilized alloc crate seems to say that we should prefer that pattern:

https://rust-lang.github.io/rfcs/2480-liballoc.html

i.e. they say to do 

```
#![no_std]

extern crate alloc;

use alloc::vec::Vec;
```

instead of

```
#![no_std]

#[cfg(feature = "no_std")] extern crate alloc;
#[cfg(not(feature = "no_std"))] extern crate std as alloc;

use alloc::vec::Vec;
```

DoumanAsh: Ok, if RFC was stabilized in 1.36 already then we're good to use it

cbeck88:
@tarcieri i realized that my idea about moving `Buf` and `BufMut` to a smaller crate will hit a problem after this recent commit:

https://github.com/tokio-rs/bytes/commit/b6cb346adfaae89bce44bfa337652e6d218d38c4#diff-c423debf79de47ec6dd3f075e0fc9cdcR901

because now `trait Buf` has a dependency back on the `Bytes` type, which would be circular if they were two crates

I've been trying to brainstorm a simple solution to this, because I don't want to sabotage long-term goals of core-only heapless version of this crate.

I think another approach might be, instead of `buf::to_bytes()` something like

```
impl Bytes {
   ...
   fn from_buf(arg: Buf) -> Bytes {
       let mut ret = BytesMut::with_capacity(arg.remaining());
       ret.put(arg);
       ret.freeze();
       ret
    }
}
```
or some such, because then `Buf` trait doesn't need to know about `bytes`

This has the downside, though, that you don't get the benefit described in commit message:
> The advantage of `Buf::into_bytes` is that it can be optimized in cases where converting a `T: Buf` into a `Bytes` instance is efficient.

A different direction is, maybe `bytes_core` would contain simple versions of `Buf` and `BufMut` that don't have knowledge of `Bytes` type.

Then, `bytes` makes its own versions of `Buf` and `BufMut` which are defined as

```
pub trait Buf: bytes_core::Buf + Into<Bytes> {}
```

and this is what bytes exports as `Buf` and `BufMut`. The implementors of `buf` can implement `Into<Bytes>` how they want, and `Bytes` can do the efficient thing. There might be a way to make it so that the default implementation is there but the user can specialize it...

But in heapless context you wouldn't try to have `Bytes` or `Into<Bytes>` at all, you would just use the `bytes_core` versions of traits and probably do some kind of allocator-aware thing as described in earlier PR's

IDK I don't have personal need of heapless bytes in my own projects right now but I'm interested to keep brainstorming, and interested what others think: @DoumanAsh ^

My feeling is that `bytes` crate `Buf` and `BufMut` objects are much nicer than the `std::io` stuff. I worked in embedded c++ development at tesla in their autopilot project for a year, and I ended up writing some very similar stuff to Buf and BufMut, but in C++, to help them do zero-copy logging and serialization in the critical path throughout their codebase. This part of `bytes` documentation speaks volumes to me:

> At first glance, it may seem that `Buf` and `BufMut` overlap in functionality with `std::io::Read` and `std::io::Write`. However, they serve different purposes. A buffer is the value that is provided as an argument to `Read::read` and `Write::write`. `Read` and `Write` may then perform a syscall, which has the potential of failing. Operations on `Buf` and `BufMut` are infallible.

If I were ever to write a serialization lib in rust I would probably try to use `bytes` rather than the `std::io` stuff, just as `prost` did. And if I wanted it to not make dynamic memory allocations (which can be forbidden in safety-critical programming contexts), and use pre-sized arenas etc., then I guess I would probably want a core-only version of `bytes::Buf` and `BufMut`. So anyways I'm really sympathetic to this as a development goal. There should be a way to make it possible, I'm just not sure what the best way is.

tarcieri:
@cbeck88 right now the main solution we have for this particular problem in an embedded context is to used fixed-sized fields for everything in the protos, so e.g. `fixed32` for all integers, and fixed-sized `string`/`bytes`, then more or less handrolling the parser/serializer to slice the data up or write it out to a fixed-sized buffer with constant protobuf framing. It'd sure be nice to find some way to leverage `heapless` in conjunction with `prost` though, so we could have prost autogenerate the parser/serializer, even for embedded.

DoumanAsh:
@cbeck88 I'm not sure in particular what will be benefit of moving traits themself, but this particular memthod could be maid through trait extension rather than moved. i.e. `BufExt` should be in the same crate as `Bytes`

cbeck88:
@DoumanAsh it's only to try to separate out the ideas and reduce the need for features to configure the package

it might be simpler to just say, there will be an `std` feature, and an `alloc` feature, and when neither are on we try to provide `Buf` and `BufMut` without `Bytes` type or related api

if we think that sounds better, it might be a good idea to add an `alloc` feature, even if it doesn't do anything right now, so that people can select it and it won't be a breaking change later if we make the core-only config work

DoumanAsh:
@cbeck88 I believe right now it should be quite simple to make the crate no_std compatible.
I don't really think we'll need a need for `alloc` feature because even if we link with it, it is not necessary to use it as long as you're not touch allocations.

I don't really see much harm in splitting but I'm myself unsure of whether we should fragment such small crate, I'd suggest to hop onto tokio gitter and discuss it

PR #283, opened 2019-08-15
Suppress deprecated warnings

taiki-e:
[`mem::uninitialized()`](https://doc.rust-lang.org/nightly/core/mem/fn.uninitialized.html) will be deprecated in 1.39.


taiki-e:
Perhaps it was almost covered by #278

PR #293, opened 2019-09-08
Fix build fail with serde feature

taiki-e:
And added check to CI.

Also, fix build fail with --no-default-features (https://github.com/tokio-rs/bytes/pull/293/commits/bbf7785c722266d566a01d7fb8a099cb44e3bef5)

### DIFF
--- a/ci/azure-test-stable.yml
+++ b/ci/azure-test-stable.yml
@@ -46,5 +46,5 @@ jobs:
     displayName: cargo ${{ parameters.cmd }} --all-features
 
   # Run with no default features
-  - script: cargo ${{ parameters.cmd }} --no-default-features
-    displayName: cargo ${{ parameters.cmd }} --no-default-features
+  - script: cargo check --no-default-features
+    displayName: cargo check --no-default-features
--- a/ci/azure-test-stable.yml
+++ b/ci/azure-test-stable.yml
@@ -29,7 +29,7 @@ jobs:
     displayName: cargo ${{ parameters.cmd }}
 
   # Run with each specified feature
-  - ${{ each feature in parameters.features.value }}:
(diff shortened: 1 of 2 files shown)


ci/azure-test-stable.yml
```diff
@@ -41,6 +41,10 @@ jobs:
...
-  # Run with no default defaults
-  - script: cargo ${{ parameters.cmd }}
-    displayName: cargo ${{ parameters.cmd }} --no-default-features
+  # Run with all features
+  - script: cargo ${{ parameters.cmd }} --all-features
+    displayName: cargo ${{ parameters.cmd }} --all-features
+
+  # Run with no default features
+  - script: cargo check --no-default-features
+    displayName: cargo check --no-default-features
```
DoumanAsh: Should still remain `${{ parameters.cmd }}`?

taiki-e: [I changed it to check only](https://github.com/tokio-rs/bytes/pull/293/commits/bbf7785c722266d566a01d7fb8a099cb44e3bef5) because some (many?) tests seemed to depend on the std feature.

taiki-e: I think we can change this to `${{ parameters.cmd }}`, but as bytes doesn't change the behavior between std and no_std, check-only should be fine.

(the actual script that is executed on CI is `script:`, not `displayName:`, so the current master has not tested/checked with no-std.)

https://github.com/tokio-rs/bytes/blob/97c13388331c6e2948278d73d3644b07a3946dda/ci/azure-test-stable.yml#L45-L46

DoumanAsh: Ah yeah indeed.

DoumanAsh:
Thanks

PR #297, opened 2019-10-04
adding impl for std::ops::BitXor

necrobious:
An impl of bit-wise xor of two Bytes, returning None if both Bytes are not of the same length.


carllerche:
Thanks for submitting this.

Could you provide the motivation. Is there prior art for bitwise ops implemented on slices in `std`?

necrobious:
Great question. To be honest, I'm not certain. I'm fairly new to rust, so still learning my way around. I was working with a crate that was using Bytes, and wanted a simple way to xor two sha256 slices together (an [example](https://github.com/necrobious/json_digest/blob/master/src/sha256.rs) of this, using slices) . I'm a new user to Bytes, so i took the opportunity to add this update for my project, and learn a little bit about Bytes in the process. I opened the PR in case it was useful to others, feel free to close if it is not a good fit.
-cheers

PR #298, opened 2019-10-11
Refactor Bytes to use an internal vtable

### LINKED ISSUE #294: Consider switching `Bytes` to a "trait object", supporting custom memory management strategies.
Currently, `Bytes` uses its own internal implementation for memory management. It supports a few strategies:

- `Vec<u8>`
- `Arc<Vec<u8>>`
- &'static [u8]
- an inlined strategy.

However, using a fixed strategy prohibits alternate strategies that would be more appropriate to the use case in question. One example is initializing a `Bytes` instance backed by a file using `mmap`.

## Proposal

The `Bytes` struct would be updated to:

```rust
struct Bytes {
    raw: RawBytes,
}

struct RawBytes {
    ptr: *const u8,
    len: usize,
    data: NonNull<()>,
    vtable: &'static Vtable,
}

struct Vtable {
    clone: unsafe fn(ptr: *const u8, len: usize, data: NonNull<()>) -> Bytes;
    drop: unsafe fn(data: NonNull<()>);
}
```

An additional `Bytes::from_raw` constructor would be added. The rest of the `Bytes` API can be implemented using the above APIs.

## Provided strategies

By default, the `bytes` crate should provide some strategies for using `Bytes`. A default of `Arc<Vec<u8>>` seems sufficient. Feature flags can be used to opt-out of the default implementation.

## `BytesMut`

The `BytesMut` structure is no longer a core part of the `bytes` API. It is unclear if it should stay in the `bytes` crate as an option or move out to a separate crate.

## Inline representation

This proposal would remove the ability to initialize a `Bytes` instance using the "inline" representation. It is unclear how big of an impact this would have. It is also unclear if use cases that take advantage of "inline" representation would be able to use an alternate strategy enabled with this change.

Refs #269

seanmonstar:
Bytes is a useful tool for managing multiple slices into the same region
of memory, and the other things it used to have been removed to reduce
complexity. The exact strategy for managing the multiple references is
no longer hard-coded, but instead backing by a customizable vtable.

- Removed ability to mutate the underlying memory from the `Bytes` type.
- Removed the "inline" (SBO) mechanism in `Bytes`. The reduces a large
  amount of complexity, and improves performance when accessing the
  slice of bytes, since a branch is no longer needed to check if the
  data is inline.
- Removed `Bytes` knowledge of `BytesMut` (`BytesMut` may grow that
  knowledge back at a future point.)

Closes #294

### DIFF
--- a/src/buf/chain.rs
+++ b/src/buf/chain.rs
@@ -186,14 +186,6 @@ impl<T, U> Buf for Chain<T, U>
         n += self.b.bytes_vectored(&mut dst[n..]);
         n
     }
-
-    fn to_bytes(&mut self) -> crate::Bytes {
-        let mut bytes: crate::BytesMut = self.a.to_bytes().try_mut()
-            .unwrap_or_else(|bytes| bytes.into());
-
-        bytes.put(&mut self.b);
-        bytes.freeze()
-    }
 }
 
 impl<T, U> BufMut for Chain<T, U>
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1,14 +1,13 @@
-use crate::{Buf, BufMut};
+use core::{cmp, fmt, hash, mem, ptr, slice, usize};
+use core::iter::{FromIterator};
+use core::ops::{Deref, RangeBounds};
+
+use alloc::{vec::Vec, string::String, boxed::Box, borrow::Borrow};
+
+use crate::Buf;
 use crate::buf::IntoIter;
 use crate::debug;
-
-use core::{cmp, fmt, mem, hash, slice, ptr, usize};
-use core::ops::{Deref, DerefMut, RangeBounds};
-use core::sync::atomic::{self, AtomicUsize, AtomicPtr};
-use core::sync::atomic::Ordering::{Relaxed, Acquire, Release, AcqRel};
-use core::iter::{FromIterator, Iterator};
-
-use alloc::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
+use crate::loom::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
 
 /// A reference counted contiguous slice of memory.
 ///
@@ -24,36 +23,22 @@ use alloc::{vec::Vec, string::String, boxed::Box, borrow::{Borrow, BorrowMut}};
 /// ```
 /// use bytes::Bytes;
 ///
-/// let mut mem = Bytes::from(&b"Hello world"[..]);
+/// let mut mem = Bytes::from("Hello world");
 /// let a = mem.slice(0..5);
 ///
-/// assert_eq!(&a[..], b"Hello");
+/// assert_eq!(a, "Hello");
 ///
 /// let b = mem.split_to(6);
 ///
-/// assert_eq!(&mem[..], b"world");
-/// assert_eq!(&b[..], b"Hello ");
+/// assert_eq!(mem, "world");
+/// assert_eq!(b, "Hello ");
 /// ```
 ///
 /// # Memory layout
 ///
-/// The `Bytes` struct itself is fairly small, limited to a pointer to the
-/// memory and 4 `usize` fields used to track information about which segment of
-/// the underlying memory the `Bytes` handle has access to.
-///
-/// The memory layout looks like this:
-///
-/// ```text
-/// +-------+
-/// | Bytes |
-/// +-------+
-///  /      \_____
-/// |              \
-/// v               v
-/// +-----+------------------------------------+
-/// | Arc |         |      Data     |          |
-/// +-----+------------------------------------+
-/// ```
+/// The `Bytes` struct itself is fairly small, limited to 4 `usize` fields used
+/// to track information about which segment of the underlying memory the
+/// `Bytes` handle has access to.
 ///
 /// `Bytes` keeps both a pointer to the shared `Arc` containing the full memory
 /// slice and a pointer to the start of the region visible by the handle.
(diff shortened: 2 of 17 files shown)


seanmonstar:
Note, I didn't actually make the vtable public yet...

seanmonstar:
The benchmarks (below) are promising! They show that `deref`, the most common operation by far, is about twice as fast. Even cloning and dropping is a little faster in most cases.

The main regression is when the data would previously have fit in the "inline" variant. The assumption is that the likelihood of data fitting inline is sufficiently low compared to the amount of times we do the other operations.

## Before

```
running 20 tests
test alloc_big                      ... bench:          67 ns/iter (+/- 1)
test alloc_mid                      ... bench:          62 ns/iter (+/- 0)                                                                                                                                                                      test alloc_small                    ... bench:       3,318 ns/iter (+/- 61)
test alloc_write_split_to_mid       ... bench:         114 ns/iter (+/- 1)
test clone_arc                      ... bench:      28,577 ns/iter (+/- 220)
test clone_inline                   ... bench:       5,366 ns/iter (+/- 477)
test clone_static                   ... bench:       5,281 ns/iter (+/- 88)
test deref_inline                   ... bench:       1,119 ns/iter (+/- 44)
test deref_shared                   ... bench:         979 ns/iter (+/- 40)
test deref_two                      ... bench:       1,236 ns/iter (+/- 79)
test deref_unique                   ... bench:         980 ns/iter (+/- 46)                                                                                                                                                                     test deref_unique_unroll            ... bench:       1,054 ns/iter (+/- 25)
test drain_write_drain              ... bench:         650 ns/iter (+/- 20)
test fmt_write                      ... bench:          36 ns/iter (+/- 1) = 1027 MB/s
test from_long_slice                ... bench:          82 ns/iter (+/- 3) = 1560 MB/s
test slice_avg_le_inline_from_arc   ... bench:      24,499 ns/iter (+/- 363)
test slice_empty                    ... bench:      19,038 ns/iter (+/- 461)
test slice_large_le_inline_from_arc ... bench:      23,458 ns/iter (+/- 621)
test slice_short_from_arc           ... bench:      22,943 ns/iter (+/- 367)                                                                                                                                                                    test split_off_and_drop             ... bench:     113,326 ns/iter (+/- 5,070)
```

## After

`bytes.rs`:
```
test clone_arc_vec                  ... bench:      18,692 ns/iter (+/- 542)
test clone_shared                   ... bench:      21,608 ns/iter (+/- 864)
test clone_static                   ... bench:       4,590 ns/iter (+/- 281)
test deref_shared                   ... bench:         425 ns/iter (+/- 9)
test deref_static                   ... bench:         426 ns/iter (+/- 18)
test deref_unique                   ... bench:         427 ns/iter (+/- 13)
test from_long_slice                ... bench:          86 ns/iter (+/- 5) = 1488 MB/s
test slice_avg_le_inline_from_arc   ... bench:      33,667 ns/iter (+/- 2,360)
test slice_empty                    ... bench:       4,371 ns/iter (+/- 133)
test slice_large_le_inline_from_arc ... bench:      35,964 ns/iter (+/- 1,015)
test slice_short_from_arc           ... bench:      33,599 ns/iter (+/- 1,455)
test split_off_and_drop             ... bench:     122,869 ns/iter (+/- 5,183)
```

`bytes_mut.rs`:
```
test alloc_big                ... bench:          70 ns/iter (+/- 1)
test alloc_mid                ... bench:          63 ns/iter (+/- 1)
test alloc_small              ... bench:      39,019 ns/iter (+/- 2,061)
test alloc_write_split_to_mid ... bench:         115 ns/iter (+/- 5)
test clone_frozen             ... bench:      21,638 ns/iter (+/- 688)
test deref_shared             ... bench:         568 ns/iter (+/- 85)
test deref_two                ... bench:         426 ns/iter (+/- 7)
test deref_unique             ... bench:         599 ns/iter (+/- 138)
test deref_unique_unroll      ... bench:         442 ns/iter (+/- 23)
test drain_write_drain        ... bench:         582 ns/iter (+/- 24)
test fmt_write                ... bench:          28 ns/iter (+/- 0) = 1321 MB/s
```

src/buf/chain.rs
```diff
@@ -187,13 +187,15 @@ impl<T, U> Buf for Chain<T, U>
         n
     }
 
+    /*
```
carllerche: I assume we either want to delete this or re-implement it.

carllerche:
> The assumption is that the likelihood of data fitting inline is sufficiently low compared to the amount of times we do the other operations.

Also, the added flexibility from the dynamism will allow handling those cases using other strategies.

src/bytes.rs
```diff
@@ -483,26 +150,13 @@ impl Bytes {
...
-    ///
-    /// assert!(Bytes::with_capacity(4).is_inline());
-    /// assert!(!Bytes::from(Vec::with_capacity(4)).is_inline());
-    /// assert!(!Bytes::with_capacity(1024).is_inline());
-    /// ```
-    pub fn is_inline(&self) -> bool {
-        self.inner.is_inline()
-    }
 
     ///Creates `Bytes` instance from slice, by copying it.
```
carllerche:
```suggestion
    /// Creates `Bytes` instance from slice, by copying it.
```

src/bytes.rs
```diff
@@ -871,2141 +449,453 @@ impl Buf for Bytes {
...
-    fn eq(&self, other: &[u8]) -> bool {
-        self.inner.as_ref() == other
-    }
-}
+impl From<Vec<u8>> for Bytes {
+    fn from(vec: Vec<u8>) -> Bytes {
+        let slice = vec.into_boxed_slice();
+        let len = slice.len();
+        let ptr = slice.as_ptr();
+        drop(Box::into_raw(slice));
```
carllerche: NIT: it might be clearer to do:

```suggestion
        let ptr = Box::into_raw(slice);
```

carllerche: It looks like multiline suggestions dont work yet.

It might be clearer if you combine these two lines into one:

```rust
let ptr = Box::into_raw(slice);
```

src/bytes.rs
```diff
@@ -871,2141 +449,453 @@ impl Buf for Bytes {
...
 
-impl PartialOrd<Vec<u8>> for Bytes {
-    fn partial_cmp(&self, other: &Vec<u8>) -> Option<cmp::Ordering> {
-        self.inner.as_ref().partial_cmp(&other[..])
-    }
-}
+// ===== impl SharedVtable =====
 
-impl PartialEq<Bytes> for Vec<u8> {
-    f
```
carllerche: I think `Relaxed` is sufficient here.

carllerche: I think that this can be `compare_exchange` with `Acquire` ordering on failure.

carllerche: It might be useful to pass in `&mut AtomicPtr` here. This allows `get_mut()` instead of doing an atomic load.

src/bytes_mut.rs
```diff
@@ -0,0 +1,1427 @@
...
+
+fn original_capacity_from_repr(repr: usize) -> usize {
+    if repr == 0 {
+        return 0;
+    }
+
+    1 << (repr + (MIN_ORIGINAL_CAPACITY_WIDTH - 1))
+}
+
+/*
```
carllerche: Delete or uncomment.

seanmonstar:
Ah woops, loom's `AtomicPtr` doesn't yet have `get_mut`, PR up: https://github.com/carllerche/loom/pull/80

carllerche:
Loom release coming: https://github.com/carllerche/loom/pull/81

src/bytes.rs
seanmonstar: That doesn't work, since it returns `*mut [u8]`, not `*mut u8`.

src/bytes.rs
```diff
@@ -705,8 +360,13 @@ impl Bytes {
     /// ```
     ///
     /// [`split_off`]: #method.split_off
+    #[inline]
     pub fn truncate(&mut self, len: usize) {
-        self.inner.truncate(len);
+        if len >= self.len {
+            self.len = 0;
```
mzabaluev: I don't think this is correct. The documentation says `truncate` should have no effect in this case.

PR #302, opened 2019-10-16
implement Buf for IoSlice and BufMut for IoSliceMut

seanmonstar:
Unsafe used to access the correct lifetimes :sob: https://github.com/rust-lang/rust/issues/62726#issuecomment-542826827


seanmonstar:
I'm undecided if this should be added or not...

carllerche:
@seanmonstar can you list out the pros / cons?

seanmonstar:
Meh, we can always add it in the future.

PR #306, opened 2019-10-30
Move "extra" methods to extension traits

seanmonstar:


src/buf/mod.rs
```diff
@@ -18,25 +18,14 @@
...
 pub use self::buf_impl::Buf;
 pub use self::buf_mut::BufMut;
+pub use self::ext::{BufExt, BufMutExt};
 #[cfg(feature = "std")]
 pub use self::buf_mut::IoSliceMut;
-pub use self::chain::Chain;
 pub use self::iter::IntoIter;
-#[cfg(feature = "std")]
-pub use self::reader::Reader;
-pub use self::take::Take;
```
ComputerDruid: AFAICT this unexports Take, which prevents library users from naming the type returned by `BufExt::take()`. Was this intentional?

Example use in h2:
https://github.com/hyperium/h2/blob/fac165e451460a787d1431d4c48af1ea2f589bfc/src/proto/streams/prioritize.rs#L68

carllerche: `Take` should be public.

seanmonstar: It was done so in https://github.com/tokio-rs/bytes/pull/318.

PR #311, opened 2019-11-09
Make Bytes::from_static a const fn?

andrewtj:
I think it'd be useful to make `Bytes::from_static` a const fn so that `Bytes` and types built on `Bytes` can be used in consts and statics. This would require a minimum Rust version of 1.39 though. (Opened a PR to discuss this instead of an issue because it's a two line change.)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -20,7 +20,7 @@ jobs:
 - template: ci/azure-test-stable.yml
   parameters:
     name: minrust
-    rust_version: 1.36.0
+    rust_version: 1.39.0
     cmd: check
 
 # Stable
@@ -37,7 +37,7 @@ jobs:
   parameters:
     name: nightly
     # Pin nightly to avoid being impacted by breakage
-    rust_version: nightly-2019-07-17
+    rust_version: nightly-2019-09-25
     benches: true
 
 # Run tests on some extra platforms
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -114,6 +114,17 @@ impl Bytes {
     /// assert_eq!(&b[..], b"hello");
     /// ```
     #[inline]
+    #[cfg(not(all(loom, test)))]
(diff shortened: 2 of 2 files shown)


andrewtj:
Oops, should have run the test suite. It's not as small a change as I thought:

```
error[E0723]: can only call other `const fn` within a `const fn`, but `const loom::sync::atomic::ptr::AtomicPtr::<()>::new` is not stable as `const fn`
   --> tests/../src/bytes.rs:121:19
    |
121 |             data: AtomicPtr::new(ptr::null_mut()),
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: for more information, see issue https://github.com/rust-lang/rust/issues/57563
    = help: add `#![feature(const_fn)]` to the crate attributes to enable

error[E0716]: temporary value dropped while borrowed
   --> tests/../src/bytes.rs:122:22
    |
122 |             vtable: &STATIC_VTABLE,
    |                     -^^^^^^^^^^^^^
    |                     ||
    |                     |creates a temporary which is freed while still in use
    |                     requires that borrow lasts for `'static`
123 |         }
124 |     }
    |     - temporary value is freed at the end of this statement

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0716, E0723.
For more information about an error, try `rustc --explain E0716`.
error: could not compile `bytes`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

carllerche:
Maybe there is a way to scope out `loom`?

seanmonstar:
If we were to switch the loom support to use `cfg(loom)`, it'd be easier to fix this.

carllerche:
@seanmonstar that is what we are moving to in Tokio due to edge cases. We probably will want to do that here as well.

seanmonstar:
Rebasing on master could allow you to only make this a `const fn` when `cfg(not(all(loom, test)))`.

andrewtj:
Thanks. I've made that change and also bumped Rust versions.

PR #316, opened 2019-11-20
implicitly grow BytesMut; add BufMutExt::chain_mut

### LINKED ISSUE #170: BytesMut should implicitly grow the internal storage in its `BufMut` implementation.
This will bring the implementation in line with that of `Vec`. This is a breaking change.

Relates to #131, #77.

carllerche:
This brings `BytesMut` in line with `Vec<u8>` behavior.

**Note** this PR also fixes an existing bug in `BytesMut::bytes_mut` that exposes invalid slices.

In order to fix a test, `BufMutExt::chain_mut` is provided. Withou this,
it is not possible to chain two `&mut [u8]`.

Closes #170

### DIFF
--- a/src/buf/ext/mod.rs
+++ b/src/buf/ext/mod.rs
@@ -124,6 +124,32 @@ pub trait BufMutExt: BufMut {
     fn writer(self) -> Writer<Self> where Self: Sized {
         writer::new(self)
     }
+
+    /// Creates an adaptor which will chain this buffer with another.
+    ///
+    /// The returned `BufMut` instance will first write to all bytes from
+    /// `self`. Afterwards, it will write to `next`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use bytes::{BufMut, buf::BufMutExt};
+    ///
+    /// let mut a = [0u8; 5];
+    /// let mut b = [0u8; 6];
+    ///
+    /// let mut chain = (&mut a[..]).chain_mut(&mut b[..]);
+    ///
+    /// chain.put_slice(b"hello world");
+    ///
+    /// assert_eq!(&a[..], b"hello");
+    /// assert_eq!(&b[..], b" world");
+    /// ```
+    fn chain_mut<U: BufMut>(self, next: U) -> Chain<Self, U>
+        where Self: Sized
+    {
+        Chain::new(self, next)
+    }
 }
 
 impl<B: BufMut + ?Sized> BufMutExt for B {}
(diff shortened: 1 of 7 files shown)


src/buf/ext/mod.rs
```diff
@@ -124,6 +124,32 @@ pub trait BufMutExt: BufMut {
     fn writer(self) -> Writer<Self> where Self: Sized {
         writer::new(self)
     }
+
+    /// Creates an adaptor which will chain this buffer with another.
```
hawkw: Super nitpicky, but IIRC this spelling of "adaptor" is the British English spelling. Elsewhere in the docs, we use American English spellings, so I think it should be:
```suggestion
    /// Creates an adapter which will chain this buffer with another.
```

src/bytes_mut.rs
```diff
@@ -924,20 +924,24 @@ impl Buf for BytesMut {
...
-        assert!(new_len <= self.cap);
+        assert!(new_len <= self.cap, "new_len = {}; capacity = {}", new_len, self.cap);
         self.len = new_len;
     }
 
     #[inline]
     fn bytes_mut(&mut self) -> &mut [mem::MaybeUninit<u8>] {
+        if self.capacity() == self.len() {
+            self.reserve(64);
+        }
```
hawkw: Does there need to be a check here that `self.capacity() != usize::MAX`, to avoid an overflow?

src/bytes_mut.rs
```diff
@@ -924,20 +924,24 @@ impl Buf for BytesMut {
...
+
         unsafe {
-            slice::from_raw_parts_mut(self.ptr.as_ptr().offset(self.len as isize) as *mut mem::MaybeUninit<u8>, self.cap)
+            slice::from_raw_parts_mut(self.ptr.as_ptr().offset(self.len as isize) as *mut mem::MaybeUninit<u8>, self.cap - self.len)
```
hawkw: Nit/TIOLI: this expression is quite long, maybe there should be a `let`? Something like
```rust
            let ptr = self.ptr.as_ptr().offset(self.len as isize) as *mut mem::MaybeUninit<u8>;
            slice::from_raw_parts_mut(ptr, self.cap - self.len)
```

carllerche:
@hawkw the bug was only on master for a bit (never released).

src/bytes_mut.rs
carllerche: If there is a need, it would be in the `reserve` function.

src/bytes_mut.rs
carllerche: Done.

src/bytes_mut.rs
carllerche: I checked and there were issues. I have resolved them and added a test.

PR #319, opened 2019-11-25
prepare v0.5.0 release

carllerche:


seanmonstar:
I'd suggest we include https://github.com/tokio-rs/bytes/pull/318 in this.

PR #335, opened 2019-12-01
impl `Zeroize` for `BytesMut` (optional)

tarcieri:
`zeroize` is a popular Rust crate for securely zeroing memory:

https://crates.io/crates/zeroize

(I am the author)

Presently it's possible to use `Zeroize` with `BytesMut` via `DerefMut` coercion (`Zeroize` is impl'd for `slice::IterMut`).

The implementation in this PR is more complete in that it handles clearing data in the remaining capacity, in the event that it previously contained secrets.

`zeroize` presently has a `bytes-preview` feature (which depends on `bytes` 4.0) which provides this implementation:

https://docs.rs/zeroize/1.0.0/zeroize/#bytes-preview-feature-zeroize-support-for-bytesmut

I'd like to either upstream that implementation, i.e. this PR, and/or drop the `bytes-preview` feature in `zeroize` as otherwise it seems deref coercion is sufficient to get a "good enough" implementation (and anyone who wants something better can make a newtype which does what's in the impl in this commit, in the event this PR doesn't get merged).

tl;dr: this PR provides an optimal implementation of `Zeroize`, but if adding an optional dependency is too onerous, no worries - what we get through deref coercion is "good enough"

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ std = []
 
 [dependencies]
 serde = { version = "1.0", optional = true }
+zeroize = { version = "1.0", optional = true, default-features = false }
 
 [dev-dependencies]
 loom = "0.2.10"
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -1392,6 +1392,16 @@ impl PartialEq<Bytes> for BytesMut {
     }
 }
 
+#[cfg(feature = "zeroize")]
+impl zeroize::Zeroize for BytesMut {
+    fn zeroize(&mut self) {
+        self.resize(self.capacity(), 0);
+        self.as_mut().zeroize();
+        debug_assert!(self.iter().all(|b| *b == 0));
+        self.clear();
+    }
+}
+
 fn vptr(ptr: *mut u8) -> NonNull<u8> {
     if cfg!(debug_assertions) {
         NonNull::new(ptr).expect("Vec pointer should be non-null")
--- a/tests/test_buf_mut.rs
+++ b/tests/test_buf_mut.rs
@@ -87,3 +87,13 @@ fn test_mut_slice() {
     let mut s = &mut v[..];
     s.put_u32(42);
 }
+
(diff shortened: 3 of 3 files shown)


seanmonstar:
My gut feeling is to reduce the 3rd party dependencies. @carllerche ?

> I'd like to either upstream that implementation, i.e. this PR, and/or drop the bytes-preview feature in zeroize as otherwise it seems deref coercion is sufficient to get a "good enough" implementation

That makes sense from a maintenance point of view. 👍

tarcieri:
@seanmonstar one semi-related thing which I don't think can be solved without `bytes` leveraging `zeroize` as an optional dependency is a `Zeroize` impl for `Bytes`. Previously we used this strategy for `SecretBytes`:

https://github.com/iqlusioninc/crates/blob/544abcc/secrecy/src/bytes.rs#L57

...or in brief: in a `Drop` impl for `(Secret)Bytes`, check if the reference count is now 0 (previously by invoking `try_mut`), and if we are, zero out the backing buffer (or if we aren't, do nothing).

However, it appears `try_mut` was removed in #298. That's probably for the best, however there's now no way to zeroize `Bytes` on drop.

I'm not sure the best way to solve this. I've thought about adding a `TryZeroize` trait which could potentially fail in these sort of cases.

carllerche:
`Bytes` is now a "trait object", though the vtable is not publicly exposed yet.

If we expose it, you should be able to define your own `SecretBytes` type and convert it to `Bytes`. Thoughts?

tarcieri:
@carllerche that might work. I'd have to play with it. What I'd like to do is be able to give out references to `&Bytes` when `expose_secret()` is called, but then ensure the backing buffer is zeroed on `Drop`:

https://docs.rs/secrecy/0.5.1/secrecy/struct.SecretBytes.html#implementations

carllerche:
Ok, I'm going to close this for now. We are trying to prune external deps (even optional) to try to reach stability.

PR #338, opened 2019-12-07
Optimize BufMut for Vec and BytesMut

seanmonstar:


src/buf/buf_mut.rs
```diff
@@ -964,6 +964,31 @@ impl BufMut for Vec<u8> {
...
+
+    fn put<T: super::Buf>(&mut self, mut src: T) where Self: Sized {
+        // In case the src isn't contiguous, reserve upfront
+        self.reserve(src.remaining());
+
+        while src.has_remaining() {
+            let l;
+
+            // a block to contain the src.bytes() borrow
+            {
```
sfackler: Nit: I think this block shouldn't be necessary with NLL

PR #341, opened 2019-12-12
Fix conversion of empty vectors to Bytes

### LINKED ISSUE #340: Bytes's Drop implementation can try to free a null pointer
This program appears to try to free a null pointer.

```rust
fn main() {
    let buf = BytesMut::with_capacity(1);
    buf.freeze();
}
```

```
#4  0x000055555555f117 in __rg_dealloc (arg0=0x0, arg1=1, arg2=1) at src/main.rs:6
#5  0x00005555555e1123 in alloc::alloc::dealloc (ptr=0x0, layout=...) at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/liballoc/alloc.rs:106
#6  0x00005555555e0cea in <alloc::alloc::Global as core::alloc::Alloc>::dealloc (self=0x7fffffffe8b0, ptr=..., layout=...) at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/liballoc/alloc.rs:177
#7  0x00005555555e1ec9 in alloc::raw_vec::RawVec<T,A>::dealloc_buffer (self=0x7fffffffe8b0) at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/liballoc/raw_vec.rs:742
#8  0x00005555555e1fdf in <alloc::raw_vec::RawVec<T,A> as core::ops::drop::Drop>::drop (self=0x7fffffffe8b0) at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/liballoc/raw_vec.rs:751
#9  0x00005555555e0b9f in core::ptr::real_drop_in_place () at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/ptr/mod.rs:175
#10 0x00005555555e0b32 in core::ptr::real_drop_in_place () at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/ptr/mod.rs:175
#11 0x00005555555e2237 in core::mem::drop (_x=<error reading variable: Cannot access memory at address 0x2>) at /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/mem/mod.rs:706
#12 0x00005555555ded38 in bytes::bytes::shared_drop (data=0x7fffffffe928, ptr=0x1 <error: Cannot access memory at address 0x1>, len=0) at /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/bytes-0.5.2/src/bytes.rs:799
#13 0x000055555555ef70 in <bytes::bytes::Bytes as core::ops::drop::Drop>::drop (self=0x7fffffffe918) at /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/bytes-0.5.2/src/bytes.rs:435
```

sfackler:
Closes #340

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -709,6 +709,13 @@ impl From<&'static str> for Bytes {
 
 impl From<Vec<u8>> for Bytes {
     fn from(vec: Vec<u8>) -> Bytes {
+        // into_boxed_slice doesn't return a heap allocation for empty vectors,
+        // so the pointer isn't aligned enough for the KIND_VEC stashing to
+        // work.
+        if vec.is_empty() {
+            return Bytes::new();
+        }
+
         let slice = vec.into_boxed_slice();
         let len = slice.len();
         let ptr = slice.as_ptr();


sfackler:
Not sure there's a great way to test this, since the default allocator doesn't actually care if you free a null pointer.

sfackler:
It seems pretty dubious to try to bitpack into the low bits of the data pointer. AFAIK there is no minimum alignment in the Rust allocator API beyond the type's alignment which is 1 in this case.

carllerche:
The vtable ptr could probably be used to disambiguate the kinds instead?

sfackler:
I seems like it's a bit tricky due to racing promotions of the same Bytes from KIND_VEC to KIND_ARC. Seems like you can maybe move the CAS from the data pointer to the vtable pointer itself though?

Alternatively, you could potentially only use `ptr` rather than `data` when freeing a KIND_VEC allocation. That way the low bit of `data` still indicates KIND_VEC vs KIND_ARC, but it's okay that it's can be a one-way transform to insert the tag bit. The KIND_ARC pointer will never have a set low bit since it does require 8 byte alignment (probably worth throwing in an assert there though), and it then doesn't matter if the KIND_VEC pointer already had the low bit set or not.

carllerche:
I mean to address the alignment issue.

IIRC the only potential race is concurrent calls to clone()

sfackler:
Yeah, if the vtable pointer is moved into an AtomicPtr, you could CAS on it instead I think.

cmcq:
For testing, MIRI gives a "type validation failed" error on the testcase in #340

davidMcneil:
Another way to test, would be to run the test suite with `jemallocator` and enable the `debug` feature flag.

RalfJung:
> It seems pretty dubious to try to bitpack into the low bits of the data pointer. AFAIK there is no minimum alignment in the Rust allocator API beyond the type's alignment which is 1 in this case.

Indeed, I think the code here is still wrong. The buffer for a `Vec<u8>` has no alignment constraints at all. Users using a custom allocator may indeed experience buffers there that are not ptr-aligned (not even 2-aligned).

seanmonstar:
Bytes now handles even and odd pointers in #346

RalfJung:
@seanmonstar I see, great. :)

PR #342, opened 2019-12-12
Assert the LSB is 0 when converting Vec into Bytes

seanmonstar:
Adding on to #341 

cc @sfackler

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -257,8 +257,20 @@ impl Bytes {
         let sub_p = subset.as_ptr() as usize;
         let sub_len = subset.len();
 
-        assert!(sub_p >= bytes_p);
-        assert!(sub_p + sub_len <= bytes_p + bytes_len);
+        assert!(
+            sub_p >= bytes_p,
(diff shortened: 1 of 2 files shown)


sfackler:
It's totally valid for an allocator to produce those pointers, though. A fix that wouldn't require a rethink of the whole implementation could be to just immediately promote odd-pointer buffers to KIND_ARC.

seanmonstar:
Yea they could, that's why I put the assert in here at least for now, so an app with a custom allocator gets a panic instead of memory unsafety.

carllerche:
How would you convert an odd pointer back to a real one if you don't know what the LSB means?

sfackler:
If you don't know what the LSB means, you can't. The idea in my comment above was to ensure that you do know what it means by avoiding the KIND_VEC case entirely if the Vec pointer is odd.

seanmonstar:
My current feeling is to merge this assertion, and perhaps fix if/when some allocator actually starts handing out odd pointers. Seem fine?

PR #344, opened 2019-12-12
Prepare v0.5.3

seanmonstar:


seanmonstar:
Shall we also yank 0.5.0, 0.5.1, and 0.5.2?

PR #346, opened 2019-12-13
Fix Bytes when Vec pointer's LSB is set

### LINKED ISSUE #343: Bytes assumes that Vec<u8>'s allocation is more than 1 byte aligned
It uses the low bit of the pointer to distinguish between KIND_VEC and KIND_ARC.

Folded over from #340

seanmonstar:
This separates the `SharedVtable` into 3:

- `PromotableEvenVtable`: The original `SharedVtable`, which will
  promote the `Vec` to `Shared` on the first clone, and is selected when
  the `Vec`'s pointer has the LSB unset.
- `PromotableOddVtable`: Similar to the `PromotableEvenVtable`, but
  selected when the `Vec`'s pointer has the LSB set. This vtable differs
  in the masking used when reconstructing the `Vec`.
- `SharedVtable`: This no longer checks if its current kind is `VEC` or
  `ARC`, and is only created by the "promotable" vtables.

This also adds a test using an "odd" global allocator that purposefully
bumps all pointers with alignment of 1.

Closes #343

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -731,20 +731,23 @@ impl From<Vec<u8>> for Bytes {
         let slice = vec.into_boxed_slice();
         let len = slice.len();
         let ptr = slice.as_ptr();
-
-        assert!(
-            ptr as usize & KIND_VEC == 0,
-            "Vec pointer should not have LSB set: {:p}",
-            ptr,
-        );
         drop(Box::into_raw(slice));
 
-        let data = ptr as usize | KIND_VEC;
-        Bytes {
-            ptr,
-            len,
-            data: AtomicPtr::new(data as *mut _),
-            vtable: &SHARED_VTABLE,
+        if ptr as usize & 0x1 == 0 {
+            let data = ptr as usize | KIND_VEC;
+            Bytes {
+                ptr,
+                len,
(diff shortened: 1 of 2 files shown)


src/bytes.rs
```diff
@@ -731,20 +731,23 @@ impl From<Vec<u8>> for Bytes {
         let slice = vec.into_boxed_slice();
```
sfackler: The is_empty special case above can be removed now that the low bit is being tracked properly.

src/bytes.rs
```diff
@@ -782,50 +785,71 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
         debug_assert_eq!(kind, KIND_VEC);
 
-        drop(rebuild_vec(shared, ptr, len));
+        drop(rebuild_vec(shared as *mut u8, ptr, len));
     }
 }
 
-unsafe fn rebuild_vec(shared: *const (), offset: *const u8, len: usize) -> Vec<u8> {
+unsafe fn rebuild_vec(buf: *mut u8, offset: *const u8, len: usize) -> Vec<u8> {
+    /*
```
sfackler: dead code

PR #377, opened 2020-03-13
Make advance_mut impl of BufMut for Vec<u8> panic if cnt > remaining

### LINKED ISSUE #354: advance_mut implementation of BufMut for Vec<u8> seems inconsistent with documentation
The documentation of `BufMut::advance_mut` states the following about panics:

> # Panics
>
> This function **may** panic if `cnt > self.remaining_mut()`.
>
> # Implementer notes
>
> It is recommended for implementations of `advance_mut` to panic if
> `cnt > self.remaining_mut()`. If the implementation does not panic,
> the call must behave as if `cnt == self.remaining_mut()`.
>
> A call with `cnt == 0` should never panic and be a no-op.

Currently, the impl for `Vec<u8>` will [reserve to accommodate the advance](https://github.com/tokio-rs/bytes/blob/master/src/buf/buf_mut.rs#L996). This seems to be in conflict with `If the implementation does not panic, the call must behave as if cnt == self.remaining_mut().`

TimHambourger:
Fixes #354.
As suggested [here](https://github.com/tokio-rs/tokio/pull/2030#discussion_r361631519), this PR goes the route of making `advance_mut` for `Vec<u8>` panic if `cnt` > `remaining`. That's definitely a breaking change: Any consumers that expect `some_vec.advance_mut(cnt)` to reserve space will need to reserve it themselves in advance instead.

### DIFF
--- a/src/buf/buf_mut.rs
+++ b/src/buf/buf_mut.rs
@@ -990,11 +990,13 @@ impl BufMut for Vec<u8> {
     unsafe fn advance_mut(&mut self, cnt: usize) {
         let len = self.len();
         let remaining = self.capacity() - len;
-        if cnt > remaining {
-            // Reserve additional capacity, and ensure that the total length
-            // will not overflow usize.
-            self.reserve(cnt);
-        }
+
+        assert!(
+            cnt <= remaining,
+            "cannot advance past `remaining_mut`: {:?} <= {:?}",
+            cnt,
+            remaining
+        );
 
         self.set_len(len + cnt);
     }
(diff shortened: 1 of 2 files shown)


carllerche:
I'm neutral on this. I don't think it is necessarily a breaking change as the fn is documented to say that panics can happen. On the other hand, the change isn't *critical* imo. The fn is already flagged as `unsafe` and callers should use caution.

So, I could go either way. @seanmonstar ?

carllerche:
Oh, if we don't merge this now, we probably should for the next breaking change.

PR #383, opened 2020-04-13
Fix reference in Take docs

fanatid:
On page https://docs.rs/bytes/0.5.4/bytes/buf/ext/struct.Take.html
`take()` lead to https://docs.rs/bytes/0.5.4/bytes/buf/ext/trait.Buf.html#method.take
> The requested resource does not exist 

Should be: https://docs.rs/bytes/0.5.4/bytes/buf/ext/trait.BufExt.html#method.take


taiki-e:
@fanatid Could you merge `tokio-rs/master` into this branch to fix CI? (looks like this PR does not allow edits by maintainers)

fanatid:
@taiki-e CI is fine now.

taiki-e:
Thanks!

PR #385, opened 2020-04-30
Allow using serde feature in no_std environment

xu-cheng:
When `serde` is used, it would enable its default feature `std`.
This of course breaks no_std build.

Fix this by disabling serde's default features. This should work for
both serde + std and serde + no_std cases.


l4l:
Any chance that the crate version is bumped? This fix is quite important imo.

taiki-e:
Published in v0.5.5

PR #392, opened 2020-05-23
Migrate CI to GitHub Actions

taiki-e:


taiki-e:
Ok, loom test failure will be fixed by https://github.com/Xudong-Huang/generator-rs/pull/20, and everything else passed.

taiki-e:
Hmm..., it seems that we need to change the "required statuses" of the branch protection settings to merge this (requires admin rights). 

@seanmonstar Could you change the branch protection settings? (I guess you have admin rights)

seanmonstar:
@taiki-e I think they're all set (the list has a lot of old options to require).

taiki-e:
@seanmonstar Thanks! I confirmed they are set. I'm going to merge this...

PR #405, opened 2020-07-02
Make Bytes & BytesMut compatible with ThreadSanitizer

tmiasko:
Replace atomic fences with atomic loads for compatibility with ThreadSanitizer.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -7,7 +7,7 @@ use alloc::{borrow::Borrow, boxed::Box, string::String, vec::Vec};
 use crate::buf::IntoIter;
 #[allow(unused)]
 use crate::loom::sync::atomic::AtomicMut;
-use crate::loom::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
+use crate::loom::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
 use crate::Buf;
 
 /// A reference counted contiguous slice of memory.
@@ -1046,7 +1046,9 @@ unsafe fn release_shared(ptr: *mut Shared) {
     // > "acquire" operation before deleting the object.
     //
     // [1]: (www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html)
-    atomic::fence(Ordering::Acquire);
+    //
+    // Use atomic load instead of fence for compatibility with ThreadSanitizer.
+    (*ptr).ref_cnt.load(Ordering::Acquire);
 
     // Drop the data
     Box::from_raw(ptr);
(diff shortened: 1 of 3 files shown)


src/bytes.rs
```diff
@@ -1046,7 +1046,7 @@ unsafe fn release_shared(ptr: *mut Shared) {
     // > "acquire" operation before deleting the object.
     //
     // [1]: (www.boost.org/doc/libs/1_55_0/doc/html/atomic/usage_examples.html)
-    atomic::fence(Ordering::Acquire);
+    (*ptr).ref_cnt.load(Ordering::Acquire);
```
seanmonstar, line 1049: This will force a second load of the `ref_cnt`, won't it?

Also, if we diverge from how `std::arc::Arc` does things, we should probably have a good reason and explain why.

tmiasko: Yes, there will be an additional load when the reference count when counter reaches zero.

I added a comment explaining why there is a load instead of a fence, so there is no need to go through git history to understand that.

To make it clear, the only reason for doing it is compatibility with ThreadSanitizer. The `std::sync::Arc` is currently implemented as proposed here, although impl is used conditionally. In overall tokio ecosystem, this is last remaining false positive I had seen reported with ThreadSanitizer.

seanmonstar: Oh I see, they've added this conditional compilation: https://github.com/rust-lang/rust/blob/f844ea1e561475e6023282ef167e76bc973773ef/src/liballoc/sync.rs#L43-L58

Could we do similar?

tmiasko: It is unstable so probably not. I don't think it is worth the complexity anyway.

On x86 this approach avoids a single mov from a location that was written a few instructions before, on a cold path that needs to do a bunch of work to deallocate the memory. Last time I looked this was essentially unmeasurable for any real world applications. For weak memory models where acquire fence leads to actual codegen, the situation is even more ambiguous.

seanmonstar: We could use `cfg(bytes_ci_tsan)` or something, setting `CARGO_CFG_BYTES_CI_TSAN=1` environment variable.

It seems it's at least worth enough that the standard library does it. Are they wrong?

tmiasko: I don't think approach from std is best idea. Requiring a custom cfg would be impractical.

seanmonstar: After reading a bit about the change in libstd (https://github.com/rust-lang/rust/pull/65097), I think we should care about the performance here, and only enable a load instead of a fence via a conditional config, to be used with tsan.

tmiasko: I don't think this is measurable outside microbenchmarks. Note that some of discussions is about different implementation.

If doing this conditionally is the only acceptable implementation, then I suggest closing this. Doing this conditionally serves no purpose, because if it doesn't work out of the box, it doesn't work period. The situation in std is different, because cfg is automatically enabled when tsan is used during compilation.

tmiasko: To clarify one point: there is no measurable impact on any benchmarks here and they do exercise this code. If you noticed anything raised your concerned I can take look again, but honestly, this change does not make any difference.

Issue #400, opened 2020-06-19
Loom should be dev dependency


JordiPolo:
Bytes brings as dependency loom which itself brings a lot of stuff. This is very surprising because loom is a testing utility.

seanmonstar:
It's listed only for `cfg(loom)`.

Ralith:
Does that actually work?

JordiPolo:
I'm getting bytes via reqwest and tokyo, no extra configuration or anything and I do get loom

jplatte:
@JordiPolo Can't reproduce. Creating a dummy project and adding `tokio` with `features = ["full"]` + `reqwest`, then running `cargo tree`, does not turn up `loom`.

JordiPolo:
Maybe is my ignorance of how Cargo works then.
Indeed `cargo tree` does not show `loom` but it is in the `Cargo.lock` file as dependency of `bytes`

jplatte:
With my dummy project it also wasn't compiled. I'm pretty sure the lockfile is target-independent, so compiling the same project for different targets doesn't change the lockfile back and forth (if it would that would be very annoying for applications that keep the lockfile in source control).

jplatte:
crates.io could really use some UI for target-specific dependencies though...

JordiPolo:
Yeah, just looking at the files in my repo I do not have enough information to know if I am going to compile this into my production application which can be scary. 
Anyways, this turned out to not be an issue, thank you for looking into it!

benma:
Please reopen this issue. It is also a problem when using `cargo vendor`, which pulls in a ton of crates that should not be there.

jplatte:
Seems like something that should be handled as an option in `cargo vendor`, like `cargo metadata`s `--filter-platform` or `cargo tree`s `--target`.

benma:
> Seems like something that should be handled as an option in `cargo vendor`, like `cargo metadata`s `--filter-platform` or `cargo tree`s `--target`.

Would be nice, but I don't think it exists? See also https://www.reddit.com/r/rust/comments/8plauq/dont_vendor_deps_i_dont_need/

jplatte:
Doesn't look like it but still, if the extra stuff being downloaded is an issue for you, you should probably try to get cargo fixed rather than having bytes work around it. The main issue seems to be https://github.com/rust-lang/cargo/issues/4544.

benma:
[...]

PR #416, opened 2020-07-13
Move loom to dev-dependencies

### LINKED ISSUE #411: loom dependency pulled in by default
Since #392 (by @taiki-e), it looks like loom went from a dev-dependency to a default dependency. It's guarded by `cfg(loom)`, but I'm not sure if that's working as intended. My application now seems to be pulling in loom as a dependency of `bytes` even though I didn't specify any `loom` cfg. As far as I understand, loom should only be used for testing?

seanmonstar:
Closes #411 #400

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,5 +30,5 @@ serde = { version = "1.0.60", optional = true, default-features = false, feature
 [dev-dependencies]
 serde_test = "1.0"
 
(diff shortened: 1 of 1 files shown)


hawkw:
I think that this might mean that `RUSTFLAGS="--cfg loom" cargo build` will fail. I don't know if that matters to us.

seanmonstar:
I don't think it does.

PR #420, opened 2020-07-20
Add inline attribute to BytesMut::split

botika:
It's just a line since `.len()` is inline

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -325,6 +325,7 @@ impl BytesMut {
     ///
     /// assert_eq!(other, b"hello world"[..]);
     /// ```
+    #[inline]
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR!

With `inline`, there is a compilation cost, so we've only added it where the method is accessed frequently enough that the function call is slowing down code. I wouldn't think `split` is called enough, but I welcome data showing it is a bottleneck.

botika:
Thanks you for quickly answer

It is for this case.
https://github.com/actix/actix-net/blob/5d6d309e66b79d837ede20a26d654274bfb68d8f/actix-codec/src/bcodec.rs#L30

Where a call would be avoided.

PR #429, opened 2020-10-16
remove vectored methods from Buf/BufMut

carllerche:
These will not be included with 1.0 for stabilization. We will explore
this later.


carllerche:
Going to do this one again.

PR #436, opened 2020-10-18
De-emphasize `Arc` implementation in `Bytes` description

Matthias247:
The previous description focussed a lot on the `Arc` based implementation
of `Bytes`. Given the vtable based implemetation, this is however not the
only valid implementation. This changes the description a bit in order
to de-emaphasize the `Arc` part, and to describe that other implementations
are possible.

This should also be necessary if the vtable gets public.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -10,7 +10,7 @@ use crate::loom::sync::atomic::AtomicMut;
 use crate::loom::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
 use crate::Buf;
 
-/// A cheeply cloneable and sliceable chunk of contiguous memory.
+/// A cheaply cloneable and sliceable chunk of contiguous memory.
 ///
 /// `Bytes` is an efficient container for storing and operating on contiguous
 /// slices of memory. It is intended for use primarily in networking code, but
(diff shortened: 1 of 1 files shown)


src/bytes.rs
```diff
@@ -10,16 +10,23 @@ use crate::loom::sync::atomic::AtomicMut;
 use crate::loom::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
 use crate::Buf;
 
-/// A reference counted contiguous slice of memory.
+/// A cheeply cloneable and sliceable chunk of contiguous memory.
```
seanmonstar, line 13: Spelling should be "cheaply".

src/bytes.rs
```diff
@@ -10,16 +10,23 @@ use crate::loom::sync::atomic::AtomicMut;
...
-/// using a reference count to track when the memory is no longer needed and can
-/// be freed.
+/// `Bytes` objects to point to the same underlying memory.
+///
+/// `Bytes` does not have a single implementation. It is an interface, whose
+/// exact behavior is implemented through dynamic dispatch in several underlying
+/// implementations of `Bytes`.
+///
+/// All `Bytes` implementations must fulfill the following requirements:
+/// - They are cheeply cloneable and thereby shareable between an unlimited amount
```
seanmonstar, line 27: "cheaply"

seanmonstar (approved):
Nice improvement!

src/bytes.rs
Matthias247, line 13:
```suggestion
/// A cheaply cloneable and sliceable chunk of contiguous memory.
```

src/bytes.rs
Matthias247, line 27:
```suggestion
/// - They are cheaply cloneable and thereby shareable between an unlimited amount
```

PR #442, opened 2020-11-09
Simplify <Vec<u8> as BufMut>::bytes_mut()

danburkert:
Removes a call to UninitMut indexing, which can panic, from
`Vec::<u8>::bytes_mut()`.  I came across this while looking at a
performance issue encountered after moving to bytes 0.6. This change
makes a microbenchmark in `prost` slightly faster, although there's
still a big delta between 0.5 and 0.6. See danburkert/prost#381.

### DIFF
--- a/src/buf/buf_mut.rs
+++ b/src/buf/buf_mut.rs
@@ -1034,15 +1034,15 @@ unsafe impl BufMut for Vec<u8> {
 
     #[inline]
     fn bytes_mut(&mut self) -> &mut UninitSlice {
-        if self.capacity() == self.len() {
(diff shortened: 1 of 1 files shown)


danburkert:
Added a second commit which adds an `#[inline]` for `Vec::put_slice`, which closed most of the remaining perf difference between 0.5 and 0.6.  Will have another PR to add inlines to UninitSlice to close out the rest.

taiki-e (approved):
LGTM, thanks!

taiki-e:
@danburkert Could you resolve merge conflicts?

PR #459, opened 2020-12-31
add inline for put_slice

laizy:
it will hurt a lot performance for methods like `BufMut::put_u8/i8` without inline. We have a test in https://github.com/sunli829/colfer-rs .

without inline,  the encode bench needs 0.612s, with inline, the time reduce to 0.362s.


Darksonn (approved):
Thanks! This seems good to me.

PR #466, opened 2021-01-20
docs: Fix broken Take link

mypetyak:
The [1.0.1 docs](https://docs.rs/bytes/1.0.1/bytes/buf/struct.Take.html) for `bytes::buf::Take` include a link to `bytes::buf::BufExt`. However, this trait no longer exists in version 1.0.1, and the link should point to the `take()` function associated with `bytes::buf::Buf`.

This PR updates the link.


taiki-e (approved):
Thanks

PR #469, opened 2021-01-22
Accept larger-than-len values for split_off and split_to, and fix tests

### LINKED ISSUE #468: Saturate `at` in `split_to` and `split_off` methods, to avoid panic
Hello,

When working with `bytes`, I noticed that `split_to` panics if the `at` value is larger than or equal to the length: https://docs.rs/bytes/1.0.1/src/bytes/bytes.rs.html#401-423

I propose that this function be modified to remove the assert, and replace line 409:
```rust
if at == self.len() {
```

to instead be:
```rust
if at >= self.len() {
```

This would mean if you have a 5-byte `Bytes` struct and you called `.split_to(10)`, it would be as if you called `.split_to(5)` on it, i.e. it would return everything that's left. 

Additionally, this makes for properly-behaving code in release mode for any given `at` value.

I believe the change would be backwards-compatible for any non-panicking code.

Are there any objections to this change? If not, I could whip up a PR.

k3d3:
This modification ended up being quicker than I expected, so I just went ahead and did it.

Let me know if there's anything else you need in this PR.

Closes #468 .

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -344,20 +344,9 @@ impl Bytes {
     /// assert_eq!(&a[..], b"hello");
     /// assert_eq!(&b[..], b" world");
     /// ```
-    ///
-    /// # Panics
-    ///
-    /// Panics if `at > len`.
     #[must_use = "consider Bytes::truncate if you don't need the other half"]
     pub fn split_off(&mut self, at: usize) -> Bytes {
-        assert!(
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for the PR! As mentioned in the issue, this would be more surprising than the current behavior, so I don't believe we should make this change.

PR #472, opened 2021-01-22
Make BytesMut::new a const fn

coolreader18:
This might depend on rust-lang/rust#65801 for the guarantee that `Vec::new()` is a dangling pointer, but in practice this works and std probably wouldn't change.

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -162,8 +162,14 @@ impl BytesMut {
     /// assert_eq!(&b"xy"[..], &bytes[..]);
     /// ```
     #[inline]
+    #[cfg(not(all(loom, test)))]
+    pub const fn new() -> BytesMut {
+        Self::from_vec_inner(NonNull::dangling(), 0, 0)
+    }
+
+    #[cfg(all(loom, test))]
     pub fn new() -> BytesMut {
-        BytesMut::with_capacity(0)
+        Self::from_vec_inner(NonNull::dangling(), 0, 0)
     }
 
     /// Returns the number of bytes contained in this `BytesMut`.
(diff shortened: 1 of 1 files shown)


coolreader18:
I also inlined functions as I was originally going to fix `BytesMut::new()` being compiled as:
```asm
foo::foo:
 push    r14
 push    rbx
 sub     rsp, 24
 mov     rbx, rdi
 mov     qword, ptr, [rsp], 1
 xorps   xmm0, xmm0
 movups  xmmword, ptr, [rsp, +, 8], xmm0
 mov     edi, 1
 call    qword, ptr, [rip, +, _ZN5bytes9bytes_mut4vptr17ha5d83dd4c2961c0aE@GOTPCREL]
 mov     r14, rax
 xor     edi, edi
 call    qword, ptr, [rip, +, _ZN5bytes9bytes_mut25original_capacity_to_repr17hedb301e0cc79d76aE@GOTPCREL]
 lea     rax, [4*rax, +, 1]
 mov     qword, ptr, [rbx], r14
 xorps   xmm0, xmm0
 movups  xmmword, ptr, [rbx, +, 8], xmm0
 mov     qword, ptr, [rbx, +, 24], rax
 mov     rax, rbx
 add     rsp, 24
 pop     rbx
 pop     r14
 ret
.LBB1_2:
 mov     rbx, rax
 mov     rdi, rsp
 call    core::ptr::drop_in_place
 mov     rdi, rbx
 call    _Unwind_Resume
 ud2
```
I'm not sure if making it const fixes that as well, but I'll leave the inlines since I think they make sense anyway.

coolreader18:
Actually, the inlines do matter lol:
```diff
diff --git a/src/bytes_mut.rs b/src/bytes_mut.rs
index 4e638c5..591b092 100644
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -1260,7 +1260,7 @@ impl Shared {
     }
 }
 
-#[inline]
+// #[inline]
 const fn original_capacity_to_repr(cap: usize) -> usize {
     let width = PTR_WIDTH - ((cap >> MIN_ORIGINAL_CAPACITY_WIDTH).leading_zeros() as usize);
     let max_capacity = MAX_ORIGINAL_CAPACITY_WIDTH - MIN_ORIGINAL_CAPACITY_WIDTH;
```
```asm
foo::foo:
 push    rbx
 mov     rbx, rdi
 xor     edi, edi
 call    qword, ptr, [rip, +, _ZN5bytes9bytes_mut25original_capacity_to_repr17hedb301e0cc79d76aE@GOTPCREL]
 lea     rax, [4*rax, +, 1]
 mov     qword, ptr, [rbx], 1
 xorps   xmm0, xmm0
 movups  xmmword, ptr, [rbx, +, 8], xmm0
 mov     qword, ptr, [rbx, +, 24], rax
 mov     rax, rbx
 pop     rbx
 ret
```

coolreader18:
Oh, maybe LTO would fix the inlining issue, since it was working when I was testing it as `foo` as a function in `bytes`, but not as a separate crate. In any case, this works too.

taiki-e (commented):
Thanks for the PR! The internal representation of BytesMut is not stable and after #435 it may be difficult to keep const so I would prefer to postpone this for now.

coolreader18:
That makes sense. Would you accept a PR to add some `#[inline]`s so that the cost of `BytesMut::new` is as if it was const?

taiki-e:
> Would you accept a PR to add some `#[inline]`s so that the cost of `BytesMut::new` is as if it was const?

I would accept a PR to do this.

PR #483, opened 2021-02-15
Override put_slice for &mut [u8]

Zettroke:
Noticed non optimal assembly generated when using `BufMut` with `&mut [u8]` slices.
[playground](https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=c6d23c716c6d1af91a242b821a7cfbce)

After further investigation add override for `put_slice` in `&mut [u8]` implementation.


seanmonstar (approved):
LGTM, thanks!

PR #498, opened 2021-05-23
Remove doc URLs

taiki-e:
- Remove html_root_url
   html_root_url is not needed in most cases. And api-guidelines no longer recommends this. see also: https://github.com/tokio-rs/tokio/pull/3489
- Remove doc URL in Cargo.toml
   https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field

    > If no URL is specified in the manifest file, crates.io will automatically link your crate to the corresponding docs.rs page.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,9 +2,7 @@
 
 name = "bytes"
 # When releasing to crates.io:
-# - Update html_root_url.
 # - Update CHANGELOG.md.
-# - Update doc URL.
 # - Create "v1.0.x" git tag.
 version = "1.0.1"
 license = "MIT"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -13,7 +11,7 @@ authors = [
     "Sean McArthur <sean@seanmonstar.com>",
 ]
 description = "Types and traits for working with bytes"
-documentation = "https://docs.rs/bytes/1.0.1/bytes/"
+documentation = "https://docs.rs/bytes"
```
taiki-e, line 14:
```suggestion
```

https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field

> If no URL is specified in the manifest file, crates.io will automatically link your crate to the corresponding docs.rs page.

Darksonn (approved):

PR #499, opened 2021-05-29
Fix a mini memory leak in Bytes on a racy promotion from vec to arc

kohensu:
Hi,
I was looking around trying to understand the implementation of `Bytes` when I found something that puzzled me in `fn shallow_clone_vec(...)`
https://github.com/tokio-rs/bytes/blob/ed1d24e57079a86c6201b5fd6be6f789265f0e6a/src/bytes.rs#L1042-L1047
It seems that this code create a memory leak (`ref_cnt: AtomicUsize::new(2)` is not deleted).
Or maybe I missed something,  in this case maybe a more explicit comment may be welcomed.

I created a PR to fix this leak.
This this just to make this code more correct - because this leak is tiny add should be rare (only when the promotion from `arc` to `vec` is performed by several threads at the same time)

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1044,7 +1044,10 @@ unsafe fn shallow_clone_vec(
             // the allocation that was made in this thread, it will not
             // be needed.
             let shared = Box::from_raw(shared);
-            mem::forget(*shared);
+            // But don't drop `_vec` as it used by the concurrent clone.
+            // Note: the partial move works here because `Shared` doesn't
+            // implement `Drop`.
+            mem::forget(shared._vec);
 
             // Buffer already promoted to shared storage, so increment ref
             // count.


carllerche:
Thanks for the PR. I will try to review early next week. However, I would recommend adding a loom test to validate the original leak and the fix. Loom (github.com/tokio-rs/loom) is able to test the rare edge conditions. I think there already are loom tests in the repo, but if you aren't able to figure it out, I (or someone else) can provide more guidance next week.

seanmonstar:
I think the code doesn't currently leak. Theres a move out of the box, which will free it. And then we forget the shared on the stack.

kohensu:
Yes, I missed the point of the move of the Shared on the stack; and because only the _vec member implement Drop everything is fine.
Sorry for the disturbance.
By the way, thanks for your quick feed back and for all the great job done on those valuable components.

Should I close this PR or you will handle this?

carllerche:
No worries, I can close it. Better to report when unsure than to ignore a mem leak :+1:

PR #502, opened 2021-07-05
Keep capacity when unsplit on empty other buf

Gbillou:
When trying to `unsplit()` from a previous `split()`:

- if the `other` buf is not empty the `unsplit()` results in a buf with the initial capacity
- if the `other` buf is empty, other is dropped, capacity lost, and anything new added to the initial buf leads to a reallocation

This changes this second behavior so that even if `other` is empty the overall capacity is kept.


Darksonn (approved):
Thanks.

PR #504, opened 2021-08-04
impl From<Box<[u8]>> for Bytes

ijackson:
Hi.

This missing impl was an inconvenience to me, and there seemed to be no reason for it not to exist.

The `From<Vec<>>` impl in fact turned the `Vec` into a `Box` almost immediately.  I moved the check for empty slices into the new `Box` impl since it is needed there too, and added a test case.

I hope this meets with your approval and thanks for your attention.


Darksonn (approved):
Thanks, this seems good to me.

PR #516, opened 2021-11-11
`const`-ify `BytesMut::len`, `BytesMut::is_empty`, and `BytesMut::capacity`,

Cyborus04:


taiki-e:
Is there a way to create BytesMut in const? If not, this does not seem to be available in practice.

taiki-e:
Ah, it was previously proposed, but I closed it for compatibility with future implementation changes. https://github.com/tokio-rs/bytes/pull/472#pullrequestreview-632946191

Cyborus04:
I guess there isn't, you're right. I think `new` could be made `const`, but I'm not sure how useful that would be.

Cyborus04:
Ah, yep, haha

PR #517, opened 2021-11-21
Add `BytesMut::zeroed`

### LINKED ISSUE #270: Support zeroed allocations (e.g `reserve_zeroed`, `with_capacity_zeroed`)
Looking at instances of `unsafe` in reqwest, there are two places ([1](https://github.com/seanmonstar/reqwest/blob/564a08f23041edb5e7384e4a4d90accdef6b06c9/src/async_impl/decoder.rs#L207), [2](https://github.com/seanmonstar/reqwest/blob/964d87ce57fabf6b27d194adf36af25e4dc0d3b2/src/body.rs#L258)) where uninitialized memory is passed to a `Read` implementation. This is a *bad idea* (as per https://github.com/rust-lang/rust/pull/62102/files, https://gankro.github.io/blah/initialize-me-maybe/) and can be fixed by just zeroing the memory beforehand.

Unfortunately `bytes` doesn't provide a good way of doing this. I imagine the ideal solution would involve something that lowers to `alloc_zeroed` / `calloc` internally.

ldm0:
This PR adds a convenient method `BytesMut::zeroed` for creating zero-filled `BytesMut`, which is more convenient than new a `BytesMut` and resize it with zero afterward. `BytesMut::zeroed` is also usually much faster.

Closes #270

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -8,6 +8,7 @@ use alloc::{
     borrow::{Borrow, BorrowMut},
     boxed::Box,
     string::String,
+    vec,
     vec::Vec,
 };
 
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Seems good to me.

PR #521, opened 2021-12-13
Add BytesMut::try_reserve()

Guiguiprim:
Add support for fallible allocation by adding `BytesMut::try_reserve()`.

This does change the MSRV to 1.57, so I don't know how we should handle it.
Adding it under a feature flag, so that people can choose not to use it ?

Fixes https://github.com/tokio-rs/bytes/issues/484


paolobarbolini:
@Guiguiprim I wouldn't say this fixes #484, as it would also be nice to get fallible allocations on the rest of the APIs

Guiguiprim:
Closing as @poliorcetics refreshed this in a new PR.

PR #523, opened 2022-01-02
Fix aliasing in Clone by using a raw pointer

saethlin:
Fixes https://github.com/tokio-rs/bytes/issues/522

I feel like it would be good to add a test case to make sure such a thing doesn't come back, but I'm not sure how to do that. There's no way to usefully test this situation without running the code under Miri, but this project doesn't fully work with Miri anyway, because Miri doesn't understand int to pointer casts.


Darksonn (approved):
Sorry, I missed this when it was posted. It looks good.

PR #527, opened 2022-01-24
Implement `Extend<Bytes>` for `BytesMut`

robjtede:
## Motivation
This lets us play nice `Stream`'s [methods](https://docs.rs/futures-util/0.3.19/futures_util/stream/trait.StreamExt.html#method.collect) like:

```rust
fn get_stream() -> impl Stream<Item = Bytes> { todo!() };
let payload = get_stream().collect::<BytesMut>().await;
```

```rust
fn get_stream() -> impl Stream<Item = Result<Bytes, E>> { todo!() };
let payload = get_stream().try_collect::<BytesMut>().await?;
```

Since the bound on this `collect` is a bit different to std's iter `collect` which is based on `FromIterator`.

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -1211,6 +1211,17 @@ impl<'a> Extend<&'a u8> for BytesMut {
     }
 }
 
+impl Extend<Bytes> for BytesMut {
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks.

Issue #503, opened 2021-07-05
Consider an unsplit method for Bytes


eloff:
I was looking for a way to join two contiguous Bytes into one, but there isn't such a thing.

It's common in parsing code to read multiple logical messages into a single BytesMut, split them off into individual Bytes for processing, but then it if they need to be written somewhere it can be advantageous to recombine them, if possible, so there are fewer syscalls to write. It's also possible to just use writev, but that typically involves an allocation for the iovec array and not all platforms support writev. It also requires much more code.

I think it is possible to do this safely:

```
impl Bytes {
    pub fn unsplit(self, other: Self) -> (Option<Self>, Option<Self>) {
        if is_contiguous(&self, &other) {
            return Some(merge(&self, &other)), None
        }
       (Some(self), Some(other))
    }
}
```

Does this seem like a worthwhile addition to the library?

eloff:
I needed this in more than one place as it turns out, so I implemented it using unsafe. Feel free to use any of this code in the bytes library, and relicense as required.

https://gist.github.com/eloff/6e0c236ae6388744a7d5b76a5e0e9b6c

xonatius:
Looks like duplicate of #287

PR #538, opened 2022-03-25
Implement Bytes::unsplit

adeschamps:
This is based largely on the example of `BytesMut::unsplit`. If two
`Bytes` are contiguous and point to the same allocation, then they are
cheaply merged. Otherwise, a new `BytesMut` is allocated, copies data
from `self` and `other`, and `self` is replaced with `new.freeze()`.

Closes https://github.com/tokio-rs/bytes/issues/503

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -422,6 +422,48 @@ impl Bytes {
         ret
     }
 
+    /// Absorbs a `Bytes` that was previously split off.
+    ///
+    /// If the two `Bytes` objects were previously contiguous, i.e.,
+    /// if `other` was created by calling `split_off` or `split_to` on
+    /// the same `BytesMut`, then this is an `O(1)` operation that
+    /// just decreases a reference count and sets a few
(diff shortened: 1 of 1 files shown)


SephVelut:
`try_unsplit` only has one fail case that's side-effect free. It doesn't pose a danger to public users. Why not make it pub?

SephVelut:
When using current `unsplit`, it's not immediately obvious to a pub user that the operation will soft-fail if the owning `BytesMut` reallocates because of insufficient capacity. In hindsight this is obvious, but this method gives off the impression that Bytes simply need to be contiguous. Emphasis on both contiguous and from the same allocation with a note in the doc that a reallocation will invalidate formerly contiguous blocks.

PonasKovas:
I also agree that there is no reason to make `try_unsplit` private.

Is there anything wrong with this PR, why are the maintainers ignoring it?

thanhminhmr:
I think this PR should also public `try_unsplit` like #746.

PR #539, opened 2022-03-25
Docs: redraw layout diagram with box drawing characters.

adeschamps:
I find this diagram very helpful, but a little hard to distinguish
between the boxes and the lines that connect them. This commit redraws
the boxes with line drawing characters so that the boxes appear a
little more solid, and stand out from the other lines.


Darksonn (approved):
Thanks.

PR #560, opened 2022-07-29
Fix reserve over allocating underlying buffer

### LINKED ISSUE #559: reserve_inner over allocates which can lead to a OOM conidition
# Summary

`reseve_inner` will double the size of the underlying shared vector instead of just extending the BytesMut buffer in place if a call to `BytesMut::reserve` would fit inside the current allocated shared buffer. If this happens repeatedly you will eventually attempt to allocate a buffer that is too large.

# Repro

```rust
fn reserve_in_arc_unique_does_not_overallocate_after_split() {
    let mut bytes = BytesMut::from(LONG);
    let orig_capacity = bytes.capacity();
    drop(bytes.split_off(LONG.len() / 2));
    let new_capacity = bytes.capacity();
    bytes.reserve(orig_capacity - new_capacity);
    assert_eq!(bytes.capacity(), orig_capacity);
}
```

schultetwin1:
Fixes calls to `reserve` when the underlying shared buffer was already
big enough to fit the requested capacity. Previously a new even larger
buffer was created anyways. This could eventually lead to an OOM
condition.

Fixes #559

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -670,7 +670,10 @@ impl BytesMut {
 
                 // Compare the condition in the `kind == KIND_VEC` case above
                 // for more details.
-                if v_capacity >= new_cap && offset >= len {
+                if v_capacity >= new_cap + offset {
+                    self.cap = new_cap;
+                    // no copy is necessary
+                } else if v_capacity >= new_cap && offset >= len {
                     // The capacity is sufficient, and copying is not too much
                     // overhead: reclaim the buffer!
 
--- a/tests/test_bytes.rs
+++ b/tests/test_bytes.rs
@@ -516,6 +516,34 @@ fn reserve_in_arc_unique_doubles() {
 }
 
 #[test]
+fn reserve_in_arc_unique_does_not_overallocate_after_split() {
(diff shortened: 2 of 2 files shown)


schultetwin1:
This appears to have been introduced by #555

schultetwin1:
Here is an example of how you could cause an OOM error

```rust
#[test]
fn reserve_in_arc_unique_does_not_overallocate_after_multiple_splits() {
    let mut bytes = BytesMut::from(LONG);
    let orig_capacity = bytes.capacity();
    for _ in 0..100 {
        drop(bytes.split_off(LONG.len() / 2));

        // now bytes is Arc and refcount == 1

        let new_capacity = bytes.capacity();
        bytes.reserve(orig_capacity - new_capacity);
    }
    assert_eq!(bytes.capacity(), orig_capacity);
}
```

Darksonn (approved):
Thanks for pointing this out. We should be careful to ensure that there aren't any other cases we missed. With this logic we can show that if `v_capacity >= 2*new_cap`, then it will not reallocate. That seems sufficient to me.

schultetwin1:
Thanks for the quick merge! And I see you are already prepping a new release (1.2.1), thank you!!

PR #577, opened 2022-11-13
Update CI toolchain to a recent nightly

Phantomical:
Currently, the loom test fails to compile some of its dependencies because the toolchain is too old. This updates the toolchain to yesterday's nightly so that CI passes.


taiki-e (approved):
Thanks.

PR #596, opened 2023-02-07
Introduce SharedBuf trait for Bytes VTable

rrichardson:
The goal of this PR is to continue the work of PR #567.  This updates the code to be more conformant to the idioms of Rust and the Tokio project.  

The work in #567 introduces a new trait which is used as a contract for internal and external structs to make themselves usable as the internal buffer for `Bytes`. 

This should solve the current issues that request a mechanism for allowing a Bytes struct to own arbitrary 3rd party buffers.  see:  https://github.com/tokio-rs/bytes/issues/437  https://github.com/tokio-rs/bytes/issues/526 https://github.com/tokio-rs/bytes/issues/571

This would allow an arbitrary buffer to be converted into a Bytes buffer (using the new Bytes::from_shared_buf method) and later be converted back to the original type, consuming the buffer (using Bytes::into_shared_buf method)
An example of this can be found [here](https://github.com/rrichardson/bytes/blob/refactor-api/tests/extern_buf_bytes.rs#L19) 

**Specific changes from #567:**
* Renames the trait `BytesImpl` into `SharedBuf`.  
* Modularizes the implementations of `SharedBuf` and moves them into their own module. 
* Changes some types to ensure that the`SharedBuf` trait can be implemented by 3rd party devs. 
* Adds an example "3rd party" `SharedBuf` implementation with tests.
* Adds a bunch of documentation. 

Due to re-organizing the SharedBuf implementations, this makes things a bit harder to diff,  but the functionality of Bytes itself is almost completely unchanged.  All tests pass, and the benchmarks are equivalent within a reasonable deviation (using `cargo benchcmp`)

For more precise deltas between the code, you can look at the commit progression in #567 

1. [change to_vec to into_vec and takes &mut AtomicPtr<()> instead of &At…](https://github.com/tokio-rs/bytes/pull/567/commits/21dd9963223ed5a493d517529788cb9a27e0b18e) 
2. [add will_truncate to Bytes vtable to prevent special casing promotabl…](https://github.com/tokio-rs/bytes/pull/567/commits/40208415abc92e124c825568a04d3504c3c1c8ba) 
3. [add Bytes::with_impl and related trait](https://github.com/tokio-rs/bytes/pull/567/commits/461c39bc3869ca286d22af69fb6ed7d071d79c07) 
4. [migrate existing vtable code to BytesImpl trait](https://github.com/tokio-rs/bytes/pull/567/commits/4231f693abcdb893716b0134aeca2292218843dd)

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -3,18 +3,14 @@ use core::iter::FromIterator;
 use core::ops::{Deref, RangeBounds};
 use core::{cmp, fmt, hash, mem, ptr, slice, usize};
 
-use alloc::{
-    alloc::{dealloc, Layout},
-    borrow::Borrow,
-    boxed::Box,
-    string::String,
-    vec::Vec,
-};
+use alloc::{borrow::Borrow, boxed::Box, string::String, vec::Vec};
 
 use crate::buf::IntoIter;
+use crate::impls::*;
 #[allow(unused)]
 use crate::loom::sync::atomic::AtomicMut;
-use crate::loom::sync::atomic::{AtomicPtr, AtomicUsize, Ordering};
+use crate::loom::sync::atomic::AtomicPtr;
+use crate::shared_buf::{BufferParts, SharedBuf};
 use crate::Buf;
 
 /// A cheaply cloneable and sliceable chunk of contiguous memory.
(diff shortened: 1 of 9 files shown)


rrichardson:
I've addressed the formatting, and will address the loom issue.  The miri issue is an interesting one, I'm having trouble divining the origin of that one.

carllerche:
Thanks for the PR. We will be reviewing it soon. Given the size, it will take a moment to get through it.

PR #629, opened 2023-09-11
Move comment to correct constant

rukai:
I could be wrong here but I think this comment is above the wrong constant.


Darksonn (approved):
Thanks.

PR #630, opened 2023-09-21
Remove ptr_map special casing for miri (#565)

shahn:
Since Rust 1.65, which introduced LLVM 15, codegen for both versions of ptr_map is the same. The release happened in Nov 2022, which should mean the majority of users have upgraded.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1221,13 +1221,6 @@ unsafe fn release_shared(ptr: *mut Shared) {
     drop(Box::from_raw(ptr));
 }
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
I wouldn't make this change until the MSRV was forced passed that version. And since this is just for tests, there isn't enough of a reason to force it.

shahn:
Bumping MSRV for sure is not warranted for this change. I made this patch due to this: https://github.com/tokio-rs/bytes/issues/565#issuecomment-1214219588

PR #641, opened 2023-12-19
Use let-else in buf_get_impl

braddunbar:
I think let-else is a bit more explicit here.


taiki-e:
Closing because let-else is not available on our MSRV.

Thanks anyway for the PR.

seanmonstar:
This would require a newer `rustc`. Even if it were within policy to pick one that supports this syntax, I wouldn't want us to do that just for a syntax change. I lean towards only upgrading MSRV when it will provide a useful improvement to users.

braddunbar:
Ah! That makes a lot of sense. Thank you for the response! 🙏

PR #642, opened 2023-12-19
Self:: -> self.

braddunbar:
I was a little confused about these calls using `Self::` instead of `self.` here. Is there a reason to use the former instead of the latter?


taiki-e (approved):
Thanks.

PR #644, opened 2023-12-22
uninit_slice: simplify as_uninit_slice_mut() logic

lucab:
This reworks `UninitSlice::as_uninit_slice_mut()` using equivalent simpler logic.


taiki-e (approved):
Thanks!

PR #652, opened 2024-01-18
BytesMut: Assert alignment of Shared

braddunbar:
Back in #362, an assertion was added to ensure that the alignment of `bytes::Shared` is even so we can use the least significant bit as a flag. `bytes_mut::Shared` uses the same technique but has no such assertion so I've added one here.

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -80,6 +80,12 @@ struct Shared {
     ref_count: AtomicUsize,
 }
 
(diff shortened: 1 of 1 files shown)


braddunbar:
Also, if this is acceptable I'd like to try abstracting things a bit to make it more clear that both sections use the same technique and how it works. As a newcomer to this codebase, I found it pretty tough to understand what was going on.

Darksonn (approved):
Thanks.

PR #662, opened 2024-01-20
Remove an unnecessary else

braddunbar:
The if block above always returns, acting as a guard, so there's no reason for an else here.


maminrayej (approved):
Make sense. Thanks.

PR #663, opened 2024-01-21
Refactor split_at/split_to

braddunbar:
I did a little refactoring of `split_at` and `split_to` to remove some extraneous checks and add some clarity around safety guarantees.

Each commit is fairly self-contained, but here are the highlights with some rationale:

1. I set `len` a little more concisely in ea9dd6050463185b8d4a9022218763902abd7c1c.
1. `split_at` and `split_to` are doing redundant bounds checking and comparison that I removed in 851b59be852b30bfa8da6b309846c37f78c41b42 and 8381ab7168ac31e2ae4d63548dcb62b6abba77b9.
1. I renamed `set_start` to `advance_unchecked` to better reflect it's usage in c70d6d2598f506a1693a08beff4e3ae6a0e23d27. I also added some explicit comments about the safety guarantees of that method and an explanation of why each usage is sound.

I'm pretty new to this repo, so if I've missed some rules or norms please let me know. I'll be happy to change things up!

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -317,7 +317,8 @@ impl BytesMut {
         );
         unsafe {
             let mut other = self.shallow_clone();
-            other.set_start(at);
+            // SAFETY: We've checked that `at` <= `self.capacity()` above.
+            other.advance_unchecked(at);
             self.cap = at;
             self.len = cmp::min(self.len, at);
             other
@@ -394,7 +395,9 @@ impl BytesMut {
             let mut other = self.shallow_clone();
             other.cap = at;
             other.len = at;
-            self.set_start(at);
+            // SAFETY: We've checked that `at` <= `self.len()` and we know that `self.len()` <=
+            // `self.capacity()`.
+            self.advance_unchecked(at);
             other
         }
     }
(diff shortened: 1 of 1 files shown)


src/bytes_mut.rs
```diff
@@ -391,8 +393,11 @@ impl BytesMut {
 
         unsafe {
             let mut other = self.shallow_clone();
-            other.set_end(at);
-            self.set_start(at);
+            other.cap = at;
+            other.len = at;
+            // SAFETY: We've checked that `at` <= `self.len()` and we know that `self.len()` <=
+            // `self.capacity()`.
+            self.advance_unchecked(at);
```
carllerche, line 400: Very minor nitpick, but this is unsafe code so I think it might be worth applying. In `split_off`, you call `advance_unchecked` first then update `cap` and `len`. You reverse it here. Would you mind keeping them the same?

carllerche (approved):
LGTM. I left a minor nit inline, but the change is an overall good improvement to clarity of the unsafe code. Thanks.

src/bytes_mut.rs
braddunbar: Yeah, thanks for pointing it out! Made things a little more consistent in b97174340bad78f2df33d8fb8bef9c841c6e511b. Let me know if that isn't what you meant.

carllerche (approved):
Thanks!

PR #665, opened 2024-01-28
Restore commented tests

braddunbar:
These seem to have been commented by accident in #298, and are still passing. All of the functionality under test here seems to still be relevant.


braddunbar:
Added one more commit to move them into a `tests` module.

carllerche (approved):
Thanks!

PR #666, opened 2024-01-28
Calculate original capacity only if necessary

braddunbar:
We don't need the original capacity if the shared data is unique, so let's not calculate it until after that check.


braddunbar:
Test failures appear to be transient, but I don't have permission to re-run them. 🤷

carllerche (approved):
Thanks.

PR #668, opened 2024-01-31
Use sub instead of offset

braddunbar:
We're always subtracting here, and we already have a usize, so `sub` seems like a more appropriate usage to me.


carllerche (approved):
Thanks! Good change.

PR #669, opened 2024-01-31
Improve the performance of put_u8 and put_i8

YHM404:
When using `put_u8/put_i8`, it internally wraps the `u8/i8` into `&[u8]`, and then calls `put_slice()`. Wrapping it into &[u8] is an unnecessary operation. So I optimized this step.

Here's the benchmark below：

**Before**
`test put_u8_bytes_mut         ... bench:         333 ns/iter (+/- 70) = 384 MB/s`

**After**
`test put_u8_bytes_mut         ... bench:         271 ns/iter (+/- 8) = 472 MB/s`

### DIFF
--- a/src/buf/buf_mut.rs
+++ b/src/buf/buf_mut.rs
@@ -319,6 +319,10 @@ pub unsafe trait BufMut {
     /// `self`.
     #[inline]
     fn put_u8(&mut self, n: u8) {
+        if self.remaining_mut() < 1 {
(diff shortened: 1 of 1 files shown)


YHM404:
> This skips some of the length checks making sure there is room for the byte. If the `Buf` is full, this would be UB.

thx. it's fixed.

YHM404:
@seanmonstar Hi, do you have any suggestions on this?

PR #670, opened 2024-01-31
get_vec_pos: use &self instead of &mut self

braddunbar:
I can't see any reason that `get_vec_pos` needs a `&mut self`. Please let me know if I'm missing something!


seanmonstar (approved):
Nice catch!

seanmonstar (approved):

PR #671, opened 2024-02-01
Copy only unread bytes in BytesMut::reserve

braddunbar:
When reserving extra space in the Vec representation, there may not be enough space to efficiently reuse the buffer. Currently, the entire buffer is copied, including bytes that have already been read. Instead, this change copies only the unread bytes to the beginning of the new buffer. As a result, fewer bytes are copied and there is more unused capacity.

It's important to note that the size of the new buffer will still be double the size of the current buffer, including bytes already read. That has not changed.

Before this change:

```
            read       unused capacity
             │           │
            ─┴─        ──┴──
old buffer: ░░░████████░░░░░

new buffer: ░░░████████░░░░░░░░░░░░░░░░░░░░░
            ─┬─        ──┬──────────────────
             │           │
            read       unused capacity
```

After this change:

```
            read       unused capacity
             │           │
            ─┴─        ──┴──
old buffer: ░░░████████░░░░░

new buffer: ████████░░░░░░░░░░░░░░░░░░░░░░░░
                    ──┬─────────────────────
                      │
                    unused capacity
```

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -629,14 +629,21 @@ impl BytesMut {
                 } else {
                     // Not enough space, or reusing might be too much overhead:
                     // allocate more space!
-                    let mut v =
-                        ManuallyDrop::new(rebuild_vec(self.ptr.as_ptr(), self.len, self.cap, off));
-                    v.reserve(additional);
+                    let request = len.checked_add(additional).expect("overflow");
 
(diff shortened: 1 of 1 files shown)


braddunbar:
Actually, this doesn't allow for extending the allocation in place via `realloc` as `reserve` does. I'm gonna think about this a little more…

PR #672, opened 2024-02-03
set_vec_pos does not need a second parameter

braddunbar:
The second argument to `set_vec_pos` always contains the value of `self.data`. Let's just use `self.data` and remove the second parameter altogether.


carllerche (approved):
Thanks

PR #674, opened 2024-03-02
Remove redundant reserve call

braddunbar:
*Note: Miri is currently failing in CI, but is fixed by https://github.com/tokio-rs/bytes/pull/673*

We're calling `reserve` twice because `put_u8` already calls it.

1. put_u8 [calls put_slice](https://github.com/tokio-rs/bytes/blob/46289278f52a26c12298779f4aaebad1dcb26d35/src/buf/buf_mut.rs#L323)
2. BytesMut::put_slice [calls BytesMut::extend_from_slice](https://github.com/tokio-rs/bytes/blob/46289278f52a26c12298779f4aaebad1dcb26d35/src/bytes_mut.rs#L1119)
3. BytesMut::extend_from_slice immediately [calls reserve](https://github.com/tokio-rs/bytes/blob/46289278f52a26c12298779f4aaebad1dcb26d35/src/bytes_mut.rs#L773)

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -1283,7 +1283,6 @@ impl Extend<u8> for BytesMut {
 
         // TODO: optimize
         // 1. If self.kind() == KIND_VEC, use Vec::extend
-        // 2. Make `reserve` inline-able
         for b in iter {
             self.put_u8(b);
         }
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -1285,7 +1285,6 @@ impl Extend<u8> for BytesMut {
         // 1. If self.kind() == KIND_VEC, use Vec::extend
         // 2. Make `reserve` inline-able
         for b in iter {
-            self.reserve(1);
             self.put_u8(b);
         }
     }


seanmonstar (approved):

src/bytes_mut.rs
```diff
@@ -1283,9 +1283,7 @@ impl Extend<u8> for BytesMut {
 
         // TODO: optimize
         // 1. If self.kind() == KIND_VEC, use Vec::extend
-        // 2. Make `reserve` inline-able
```
taiki-e, line 1286: I'm not sure what this TODO comment originally meant, but if it means something like "reserve (in here or put_u8) should be able to omit if the compiler has knowledge of the actual length of iter" it may not have been resolved yet.

([In my experience](https://github.com/openrr/mesh-loader/blob/2405ef08511c595fa8a754b08e15a0be32513d67/src/stl/mod.rs#L185-L187), the compiler optimizations around the reserve have not worked that well.)

braddunbar: Thanks @taiki-e! I think it refers to https://github.com/tokio-rs/bytes/pull/313, which was just after the introduction of the comment in https://github.com/tokio-rs/bytes/pull/305. I should've mentioned that in my commit though. Thanks for taking a look!

braddunbar:
Updated with #673 to fix miri tests!

taiki-e (approved):

Darksonn (changes requested):
Before accepting this PR, I would like to see a test that calls `extend` with an iterator containing more items than the lower limit of the size hint. I'm worried about the panic conditions of `put_u8`.

braddunbar:
@Darksonn Sure thing! Added in 76dae7b880c52384d6a3edb364f1375b45a9274f. I think the [following test](https://github.com/tokio-rs/bytes/blob/76dae7b880c52384d6a3edb364f1375b45a9274f/tests/test_bytes.rs#L623-L631) technically covers it via `Extend<&'a u8>`, but I like being explicit.

PR #675, opened 2024-03-02
Use ManuallyDrop instead of mem::forget

braddunbar:
This change converts several usages of `mem::forget` to `ManuallyDrop` as recommended [in the docs][docs] for `mem::forget`.

> While mem::forget can also be used to transfer memory ownership, doing so is error-prone. [ManuallyDrop](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html) should be used instead.

[docs]: https://doc.rust-lang.org/std/mem/fn.forget.html#relationship-with-manuallydrop

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -244,23 +244,22 @@ impl BytesMut {
     /// ```
     #[inline]
     pub fn freeze(self) -> Bytes {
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thank you.

PR #676, opened 2024-03-03
copy_to_bytes: Add panic section to docs

### LINKED ISSUE #454: Document that `copy_to_bytes()` panics if len > remaining
This is left implicit in the documentation, would be better to make this explicit, I think?

braddunbar:
Fixes #454.


taiki-e (approved):
Thanks.

PR #683, opened 2024-03-25
feat: support From<&[u8; N]> for Bytes

cjwcommuny:


braddunbar:
Hey @cjwcommuny! Out of curiosity, are you having an issue using `Bytes::copy_from_slice` directly or is this about the convenience of calling `from`/`into`?

cjwcommuny:
@braddunbar  I would like to write a function like:

```rust
fn foo(x: impl Into<Bytes>) {}

foo(b"hello");
foo(Bytes::new());
```

I believe that implement `<&[u8; N]>` for `Bytes` would improve the ergonomics.

PR #689, opened 2024-04-10
Bytes::split_to - check fast path first

braddunbar:
If `at == self.len()` then we already know `at <= self.len()`. If `at == 0`, it can't be greater than `self.len()`.


Darksonn (approved):
Thanks.

PR #699, opened 2024-04-25
Improve BytesMut::split suggestion

paolobarbolini:
Incorporates the improvement from #698 into the `must_use` suggestion of `BytesMut::split`.

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -349,7 +349,7 @@ impl BytesMut {
     ///
     /// assert_eq!(other, b"hello world"[..]);
     /// ```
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
You know, this makes me realize that `advance` could be simplified to call `clear` instead of `set_len(0)`.

paolobarbolini:
> You know, this makes me realize that `advance` could be simplified to call `clear` instead of `set_len(0)`.

Yeah I had skipped `clear` because it can't avoid the branch inside `transmute`, but I guess if that was a real problem it could be solved within `clear` itself.

PR #705, opened 2024-05-06
Offset from

braddunbar:
In #695, @Sytten added a separate version of `offset_from` for `*const u8` (thanks @Sytten). After some digging, I found that we can use the new one for both cases. Further, there are several additional places we can/should be using `offset_from` instead of subtracting directly.


Darksonn (approved):
Thank you.

PR #708, opened 2024-05-20
Implement `Buf::chunks_vectored` for `VecDeque<u8>`

paolobarbolini:
A tiny step closer at making #474 and #701 harder to encounter.


paolobarbolini:
I'm starting to think this will require something more creative to get `VecDeque` to become noncontinuous.

Darksonn (approved):
Thank you.

PR #715, opened 2024-06-19
Add optional support for the arbitrary crate

morr0ne:
This PRs adds optional support for the [arbitrary crate](https://github.com/rust-fuzz/arbitrary/) by deriving the Arbitrary trait for both Bytes and BytesMut.
This is incredibly useful for fuzz testing as it allows for [structure aware fuzzing](https://rust-fuzz.github.io/book/cargo-fuzz/structure-aware-fuzzing.html).
Adding this traits directly allows downstream users to easily derive the traits for their structs:

E.g. from one of my projects

```rust
#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Arbitrary)]
pub struct Message {
    pub object_id: u32,
    pub opcode: u16,
    pub payload: Bytes,
}
```

Without bytes implementing Arbitrary directly this would require some kinda of wrapper struct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ default = ["std"]
 std = []
 
 [dependencies]
+arbitrary = { version = "1.1.6", optional = true }
 serde = { version = "1.0.60", optional = true, default-features = false, features = ["alloc"] }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


seanmonstar:
Thanks for the PR! However, I wouldn't want to add this public dependency to bytes.

morr0ne:
> Thanks for the PR! However, I wouldn't want to add this public dependency to bytes.

What's the issue with having arbitrary as a depedency? It is optional and not enabled by default just like serde

morr0ne:
Could you please provide an update on the status of this? The changes seem straightforward and ready for merging. Are there any specific concerns or reasons I am not aware of?

PR #728, opened 2024-08-02
test: ensure BytesMut::advance reduces capacity

camshaft:
This PR adds a test to ensure the behavior of `BytesMut::advance` is preserved in order to avoid #725.

I also confirmed that this test fails with the change from #698.

```
---- advance_bytes_mut_remaining_capacity stdout ----
testing capacity=0, len=0, advance=0
testing capacity=1, len=0, advance=0
testing capacity=1, len=1, advance=0
testing capacity=1, len=1, advance=1
thread 'advance_bytes_mut_remaining_capacity' panicked at tests/test_bytes.rs:701:17:
assertion `left == right` failed: Buf::advance should reduce the remaining capacity
  left: 1
 right: 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    advance_bytes_mut_remaining_capacity
```

### DIFF
--- a/tests/test_bytes.rs
+++ b/tests/test_bytes.rs
@@ -676,6 +676,43 @@ fn advance_bytes_mut() {
     assert_eq!(a, b"d zomg wat wat"[..]);
 }
 
+// Ensures BytesMut::advance reduces always capacity
+//
+// See https://github.com/tokio-rs/bytes/issues/725
+#[test]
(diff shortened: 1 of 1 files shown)


camshaft:
Looks like my brute force test is quite slow under `miri`. I think it's probably a good idea to reduce the search space a bit under that mode.

seanmonstar (approved):

PR #731, opened 2024-08-19
Fix double spaces in comments and doc comments

paolobarbolini:
Caught by this very naive regex: `[a-zA-Z0-9]  [a-zA-Z0-9]`


Darksonn (approved):
Thank you.

PR #737, opened 2024-09-17
ci: bump nightly to 2024-09-15

seanmonstar:


paolobarbolini:
I think you missed #734

PR #742, opened 2024-10-23
feat: Bytes::from_owner

amunra:
Introduces `Bytes::from_owner`, allowing the creation of a bytes object from another `AsRef<[u8]>` object which is responsible for the ownership of the exposed byte slice.

The core use case is to safely construct from mapped memory.

```rust
use memmap2::Map;

let file = File::open("upload_bundle.tar.gz")?;
let mmap = unsafe { Mmap::map(&file) }?;
let b = Bytes::from_owner(mmap);
```

It should practically address the vast majority of use cases for wanting to expose the VTable as has often been requested in https://github.com/tokio-rs/bytes/issues/437, without actually exposing the VTable.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -212,25 +212,18 @@ impl Bytes {
     /// use memmap2::Map;
     ///
     /// let file = File::open("upload_bundle.tar.gz")?;
-    /// let b = unsafe {
-    ///     let mmap = Mmap::map(&file)?;
-    ///     let ptr = mmap.as_ptr();
-    ///     let len = mmap.len();
-    ///     Bytes::with_owner(ptr, len, mmap)
-    /// };
+    /// let mmap = unsafe { Mmap::map(&file) }?;
+    /// let b = Bytes::from_owner(mmap);
     /// ```
     ///
-    /// The function is marked `unsafe` because it requires the caller to ensure
-    /// that the memory region specified by `ptr` and `len` remains valid and
-    /// linked to the lifetime of the provided owner.
-    /// The `owner` will be transferred to the constructed [Bytes] object which
+    /// The `owner` will be transferred to the constructed [Bytes] object, which
     /// will ensure it is dropped once all remaining clones of the constructed
     /// object are dropped. The owner will then be responsible for dropping the
     /// specified region of memory as part of its [Drop] implementation.
     ///
-    /// Note that converting [Bytes] with an owner into a [BytesMut] will
-    /// always create a deep copy of the buffer into newly allocated memory.
-    pub unsafe fn with_owner<T>(ptr: *const u8, len: usize, owner: T) -> Self {
+    /// Note that converting [Bytes] constructed from an owner into a [BytesMut]
+    /// will always create a deep copy of the buffer into newly allocated memory.
+    pub fn from_owner<T: AsRef<[u8]>>(owner: T) -> Self {
         let owned = Box::into_raw(Box::new(Owned {
             lifetime: OwnedLifetime {
                 ref_cnt: AtomicUsize::new(1),
@@ -238,6 +231,14 @@ impl Bytes {
             },
             owner,
         }));
+
+        // Now that the ownership is moved to the Box its memory location is pinned.
+        // It's therefore safe to access the memory region, which will remain valid,
+        // even if the slice returned refers to memory within the owner object itself.
+        let owned_ref = unsafe { &(*owned) };
+        let buf = owned_ref.owner.as_ref();
+        let ptr = buf.as_ptr();
+        let len = buf.len();
 
         Bytes {
             ptr,
@@ -584,7 +585,7 @@ impl Bytes {
     /// and return self.
     ///
     /// This will also always fail if the buffer was constructed via
-    /// [with_owner](Bytes::with_owner).
+    /// [from_owner](Bytes::from_owner).
     ///
     /// # Examples
     ///
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -200,11 +200,33 @@ impl Bytes {
         }
     }
 
-    pub unsafe fn from_external<T>(ptr: *const u8, len: usize, owner: T) -> Self {
-        let external = Box::into_raw(Box::new(External {
-            lifetime: ExternalLifetime {
+    /// Create [Bytes] with a buffer whose lifetime is controlled
+    /// via an explicit owner.
+    ///
+    /// A common use case is creating bytes to a memory mapped file,
+    /// allowing for zero-copy construction.
+    ///
+    /// ```no_run
+    /// use bytes::Bytes;
+    /// use std::fs::File;
+    /// use memmap2::Map;
+    ///
+    /// let mut file = File::open("upload_bundle.tar.gz");
+    /// let b = unsafe {
+    ///     let mmap = Mmap::map(&file);
+    ///     let ptr = mmap.as_ptr();
+    ///     let len = mmap.len();
+    ///     Bytes::with_owner(ptr, len, mmap)
+    /// };
+    /// ```
+    ///
+    /// Converting to a vector or a [BytesMut] will create a deep copy of
+    /// the buffer.
+    pub unsafe fn with_owner<T>(ptr: *const u8, len: usize, owner: T) -> Self {
+        let owned = Box::into_raw(Box::new(Owned {
+            lifetime: OwnedLifetime {
                 ref_cnt: Default::default(),
-                drop: external_box_and_drop::<T>,
+                drop: owned_box_and_drop::<T>,
             },
             owner,
         }));
@@ -212,8 +234,8 @@ impl Bytes {
         Bytes {
             ptr,
             len,
-            data: AtomicPtr::new(external as _),
-            vtable: &EXTERNAL_VTABLE,
+            data: AtomicPtr::new(owned as _),
+            vtable: &OWNED_VTABLE,
         }
     }
 
@@ -1029,37 +1051,28 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
     // nothing to drop for &'static [u8]
 }
 
-// ===== impl ExternalVtable =====
-
-/// Manage the lifetime of a [Bytes] buffer externally.
-///
-/// The trait is marked as `unsafe` since it needs to ensure that the slice
-/// of returned by [AsRef<[u8]>](AsRef) retains a fixed address in memory,
-/// tied the lifetime of the trait implementor.
-///
-/// A common use case for implementing the [ExternalBytesOwner] is when
-/// the memory is owned by external means such as a memory mapped file.
+// ===== impl OwnedVtable =====
 
 #[repr(C)]
-struct ExternalLifetime {
+struct OwnedLifetime {
     ref_cnt: AtomicUsize,
     drop: unsafe fn(*mut ()),
 }
 
 #[repr(C)]
-struct External<T> {
-    lifetime: ExternalLifetime,
+struct Owned<T> {
+    lifetime: OwnedLifetime,
     owner: T,
 }
 
-unsafe fn external_box_and_drop<T>(ptr: *mut ()) {
-    let b: Box<External<T>> = Box::from_raw(ptr as _);
+unsafe fn owned_box_and_drop<T>(ptr: *mut ()) {
+    let b: Box<Owned<T>> = Box::from_raw(ptr as _);
     drop(b);
 }
 
-unsafe fn external_clone(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Bytes {
-    let external = data.load(Ordering::Acquire);
-    let ref_cnt = &(*external.cast::<ExternalLifetime>()).ref_cnt;
+unsafe fn owned_clone(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Bytes {
+    let owned = data.load(Ordering::Acquire);
+    let ref_cnt = &(*owned.cast::<OwnedLifetime>()).ref_cnt;
     let old_size = ref_cnt.fetch_add(1, Ordering::Relaxed);
 
     if old_size > usize::MAX >> 1 {
@@ -1069,29 +1082,29 @@ unsafe fn external_clone(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> By
     Bytes {
         ptr,
         len,
-        data: AtomicPtr::new(external as _),
-        vtable: &EXTERNAL_VTABLE
+        data: AtomicPtr::new(owned as _),
+        vtable: &OWNED_VTABLE
     }
 }
 
-unsafe fn external_to_vec(_data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Vec<u8> {
+unsafe fn owned_to_vec(_data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Vec<u8> {
     let slice = slice::from_raw_parts(ptr, len);
     slice.to_vec()
 }
 
-unsafe fn external_to_mut(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> BytesMut {
-    BytesMut::from_vec(external_to_vec(data, ptr, len))
+unsafe fn owned_to_mut(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> BytesMut {
+    BytesMut::from_vec(owned_to_vec(data, ptr, len))
 }
 
-unsafe fn external_is_unique(data: &AtomicPtr<()>) -> bool {
-    let external = data.load(Ordering::Acquire);
-    let ref_cnt = &(*external.cast::<ExternalLifetime>()).ref_cnt;
+unsafe fn owned_is_unique(data: &AtomicPtr<()>) -> bool {
+    let owned = data.load(Ordering::Acquire);
+    let ref_cnt = &(*owned.cast::<OwnedLifetime>()).ref_cnt;
     ref_cnt.load(Ordering::Relaxed) == 1
 }
 
-unsafe fn external_drop(data: &mut AtomicPtr<()>, _ptr: *const u8, _len: usize) {
-    let external = data.load(Ordering::Acquire);
-    let lifetime = &*external.cast::<ExternalLifetime>();
+unsafe fn owned_drop(data: &mut AtomicPtr<()>, _ptr: *const u8, _len: usize) {
+    let owned = data.load(Ordering::Acquire);
+    let lifetime = &*owned.cast::<OwnedLifetime>();
     let ref_cnt = &lifetime.ref_cnt;
 
     if ref_cnt.fetch_sub(1, Ordering::Release) != 1 {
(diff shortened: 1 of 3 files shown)


Darksonn:
What do you think, Sean? It is interesting that it avoids exposing the vtable like previous suggestions.

amunra:
> What do you think, Sean? It is interesting that it avoids exposing the vtable like previous suggestions.

Yes, that was my goal: https://github.com/tokio-rs/bytes/issues/437#issuecomment-2432439604

src/bytes.rs
```diff
@@ -1012,13 +1064,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+    // Since the memory's ownership is tied to an external owner
+    // it is never zero-copy to create a BytesMut.
+    false
+}
+
+unsafe fn owned_drop_impl(owned: *mut ()) {
+    let lifetime = owned.cast::<OwnedLifetime>();
+    let ref_cnt = &(*lifetime).ref_cnt;
+
+    let old_cnt = ref_cnt.fetch_sub(1, Ordering::Release);
```
Darksonn, line 1129: You must perform an acquire load on the refcount in the `old_cnt == 1` case before running the destructor. This is necessary to synchronize with other threads also dropping the refcount concurrently.

On the other hand, relaxed is sufficient on the `AtomicPtr` since we never modify the address after initialization.

amunra: hrm.. ok. Then `Ordering::Acquire` in `owned_is_unique` as well?

I always feel slightly unsure about the atomic ordering flags, pointers most appreciated here.

amunra: FWIW, I followed this code:
```rust
unsafe fn release_shared(ptr: *mut Shared) {
    // `Shared` storage... follow the drop steps from Arc.
    if (*ptr).ref_cnt.fetch_sub(1, Ordering::Release) != 1 {
        return;
    }
    
    ...
```

seanmonstar: The following comments afterwards from `release_shared` are important.

seanmonstar: You don't need an acquire for the is_unique check.

amunra: Done.

carllerche:
At a quick glance, this looks like it would be forwards compatible with exposing the vtable?

src/bytes.rs
```diff
@@ -200,6 +201,53 @@ impl Bytes {
...
+    /// ```
+    ///
+    /// The `owner` will be transferred to the constructed [Bytes] object, which
+    /// will ensure it is dropped once all remaining clones of the constructed
+    /// object are dropped. The owner will then be responsible for dropping the
+    /// specified region of memory as part of its [Drop] implementation.
+    ///
+    /// Note that converting [Bytes] constructed from an owner into a [BytesMut]
+    /// will always create a deep copy of the buffer into newly allocated memory.
+    pub fn from_owner<T: AsRef<[u8]>>(owner: T) -> Self {
```
sfackler, line 226: Doesn't this need some kind of StableDeref bound to protect against weird edge cases where the referenced region changes over time?

carllerche: The `string` crate had a similar issue: https://github.com/carllerche/string/issues/9

I do wonder if `Sync` would be sufficient here... In general, we probably do need a `Send` bound as well since `Bytes` is `Send`

sfackler: Sync wouldn't cover it since you can have a thread-safe implementation that still mutates internally.

sfackler: Hmm yeah only calling as_ref once may actually make this fine.

src/bytes.rs
```diff
@@ -200,6 +201,53 @@ impl Bytes {
...
+                drop: owned_box_and_drop::<T>,
+            },
+            owner,
+        }));
+
+        // Now that the ownership is moved to the Box its memory location is pinned.
+        // It's therefore safe to access the memory region, which will remain valid,
+        // even if the slice returned refers to memory within the owner object itself.
+        let owned_ref = unsafe { &(*owned) };
+        let buf = owned_ref.owner.as_ref();
```
paolobarbolini, line 239: If `as_ref` were to panic we'd be leaking memory here because `Box::into_raw` has been called

src/bytes.rs
Darksonn, line 226: I wonder if we may want our own trait instead of `AsRef<[u8]>` so we can add more methods in the future.

Darksonn: StableDeref is *not* necessary because we don't call `as_ref` more than once.

src/bytes.rs
amunra: Good point. I'll fix this.

src/bytes.rs
seanmonstar: There's a safe way for it to still be UB, even with only calling it once. If I pass in a form of `Arc<Mutex<_>>`, and then afterwards change the internal buffer completely (maybe a reallocation), the captured pointer could be referring to invalid memory.

sfackler: The reference we get back from the as_ref needs to be valid for the remaining lifetime of the object though - it can't be swapped out without a call to an &mut method that would ensure the initial borrow is gone.

Darksonn: No, it's sound as-is. You cannot write a safe thing where this breaks even with `Arc`/`Mutex` because if the buffer is inside the mutex, the implementer of `AsRef` can't return the buffer from `as_ref` since it borrows from the mutex guard.

As written, the `as_ref` call essentially borrows the `owner` field until we run the destructor. As long as we never make any mutable function calls on `owner` during that time, it's okay.

Darksonn, line 226: We need `T: Send` because we may run the destructor of `T` on another thread.

(`T: Sync` is not needed because we don't call `as_ref` in parallel; we just call it once and stash the pointer.)

paolobarbolini:
```rs
use std::sync::{Arc, Mutex};

struct Evil {
    within: Arc<Mutex<Vec<u8>>>,
}

impl AsRef<[u8]> for Evil {
    fn as_ref(&self) -> &[u8] {
        // error[E0515]: cannot return value referencing temporary value
        self.within.try_lock().unwrap().as_ref()
    }
}
```

paolobarbolini: It also needs `'static`, as demonstrated by running Miri on:


```rs
fn main() {
    let bytes = {
        let data = [0u8; 8];
        bytes::Bytes::from_owner(&data)
    };

    println!("{bytes:?}");
}
```

amunra: Added `Send + 'static`.

src/bytes.rs
amunra: Done.

carllerche:
Also, I want to thank @amunra for putting this PR together and moving the issue forward. I think this is a great/clever way to move `bytes` forward incrementally without figuring out all the complexities of exposing the vtable.

I will try to do a closer review of the code tomorrow and we may want to consider other names than `from_owned` (though I think it might be the best we come up with).

src/bytes.rs
paolobarbolini: Miri is not liking it: `test owned_basic ... error: Undefined Behavior: trying to retag from <5921171> for SharedReadOnly permission at alloc2203252[0x18], but that tag does not exist in the borrow stack for this location`

A solution to this is probably to:

1. Go back to `Box::into_raw(Box::new(...))`
2. Put the pointer in a guard struct that on `Drop` does `Box::from_raw`
3. `mem::forget` the guard struct just before constructing `Bytes`

Darksonn: Yeah, we must use `Box::into_raw` before calling `as_ref` for miri to be happy. You can reuse the destructor of `Bytes` to handle panics:
```rs
let owned = Box::into_raw(Box::new(Owned {
    lifetime: OwnedLifetime {
        ref_cnt: AtomicUsize::new(1),
        drop: owned_box_and_drop::<T>,
    },
    owner,
}));

let mut ret = Bytes {
    ptr: NonNull::dangling().as_ptr(),
    len: 0,
    data: AtomicPtr::new(owned.cast()),
    vtable: &OWNED_VTABLE,
};

let buf = unsafe { &*owned }.owner.as_ref();
ret.ptr = buf.as_ptr();
ret.len = buf.len();

ret
```

Darksonn: Please add a comment explaining this so that refactors don't break it in the future. Please also add a test that the value is dropped on panic.

src/bytes.rs
amunra: If I'm not mistaken the current API should guard against `Owner` exposing a thread local as well.

amunra: `<opinionated>`

A custom trait is how I had originally started implementing this, but gave up early on for a few reasons:

* Worse ergonomics for custom trait:
  * Many things already implement `AsRef<[u8]>`, so `Bytes::from_owner(mmap);` is easy and straight-forward.
  * Conversely, a custom trait would require `Bytes::from_custom(MyMmapWrapper::new(mmap))`.
* Safety:
  * `AsRef<[u8]>` can be called during construction and only once. It's a trick that makes the `from_owner` API safe.
  * Any additional APIs would - in practice - need to be called during the lifetime of bytes. As an example `trait CustomOwner { fn try_into_mut(self) -> Result<impl CustomOwnerMut, Self>; }` (unless I'm wrong) would make it into an `unsafe trait`.
  * What's missing from the current approach is zero-copy conversion to `BytesMut`.
    The implementor of a trait that that exposed this would need to be very careful here to respect the `Vec<u8>` semantics. Context: If `Bytes` is not unique, the existing code creates a newly allocated `Vec<u8>` copy of the referenced memory. Unless due care is taken - in the specific case of an `Mmap` - this would likely lead an implementor to erroneously assume that they should create an `MmapMut`. This would be wrong, especially if the mmapped section is backed by a file as the resulting behaviour would sometimes update the contents of a file and sometimes not, depending on `Bytes`'s internal refcount.
* Feature creep:
  * By all means, I think a custom trait approach might work in the future for a `fn from_custom <T: CustomOwner>` and as a good way of exposing more of the vtable functionality, but it is not where the current pain point in the API is and there are many hurdles to get past that as has been shown in previous efforts to tackle https://github.com/tokio-rs/bytes/issues/437.

The full-fat custom use case would mostly cover - as far as I can tell - mixing `Bytes` created and managed via custom memory allocators (QuestDB would certainly care!), but I suspect there's a more elegant way to handle those anyway.

`</opinionated>`

.. that said, if you guys prefer that route I'm happy to amend my PR in that direction.

tests/test_bytes.rs
```diff
@@ -1479,3 +1481,92 @@ fn split_to_empty_addr_mut() {
...
+    assert!(b2.is_unique());
+    let b3 = b2.slice(1..b2.len() - 1);
+    assert!(!b2.is_unique());
+    assert!(!b3.is_unique());
+    drop(b2);
+    assert_eq!(drop_counter.load(Ordering::Acquire), 0);
+    assert!(b3.is_unique());
+    drop(b3);
+    assert_eq!(drop_counter.load(Ordering::Acquire), 1);
+}
```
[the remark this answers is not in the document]
amunra: I'll split it up, this was born out of debugging :-)
Give me some time.. competing priorities.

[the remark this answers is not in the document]
amunra: Cleaned up.

src/bytes.rs
```diff
@@ -545,7 +613,7 @@ impl Bytes {
     /// assert_eq!(bytes.try_into_mut(), Ok(BytesMut::from(&b"hello"[..])));
     /// ```
     pub fn try_into_mut(self) -> Result<BytesMut, Bytes> {
-        if self.is_unique() {
+        if unsafe { (self.vtable.cheap_into_mut)(&self.data) } {
```
Darksonn, line 616: Should we change `is_unique` to always return `false` instead of adding this extra check? That's how `from_static` works.

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+    owner: T,
+}
+
+unsafe fn owned_box_and_drop<T>(ptr: *mut ()) {
+    let b: Box<Owned<T>> = Box::from_raw(ptr as _);
+    drop(b);
+}
+
+unsafe fn owned_clone(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Bytes {
+    let owned = data.load(Ordering::Acquire);
```
Darksonn, line 1104:
```suggestion
    let owned = data.load(Ordering::Relaxed);
```

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+}
+
+unsafe fn owned_to_vec(_data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Vec<u8> {
+    let slice = slice::from_raw_parts(ptr, len);
+    slice.to_vec()
+}
+
+unsafe fn owned_to_mut(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> BytesMut {
+    let bytes_mut = BytesMut::from_vec(owned_to_vec(data, ptr, len));
+    owned_drop_impl(data.load(Ordering::Acquire));
```
Darksonn, line 1126:
```suggestion
    owned_drop_impl(data.load(Ordering::Relaxed));
```

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+    bytes_mut
+}
+
+unsafe fn owned_is_unique(data: &AtomicPtr<()>) -> bool {
+    let owned = data.load(Ordering::Acquire);
```
Darksonn, line 1131:
```suggestion
    let owned = data.load(Ordering::Relaxed);
```

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+    let ref_cnt = &(*owned.cast::<OwnedLifetime>()).ref_cnt;
+    ref_cnt.load(Ordering::Relaxed) == 1
```
Darksonn, line 1133: Not strictly necessary for anything this crate does, but probably a good idea in case the user writes unsafe code relying on this value.
```suggestion
    ref_cnt.load(Ordering::Acquire) == 1
```

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+    let lifetime = owned.cast::<OwnedLifetime>();
+    let ref_cnt = &(*lifetime).ref_cnt;
+
+    let old_cnt = ref_cnt.fetch_sub(1, Ordering::Acquire);
+    if old_cnt != 1 {
+        return;
+    }
+
+    let drop = &(*lifetime).drop;
+    drop(owned)
```
Darksonn, line 1152: This is too confusing. There's already a function called `drop` in scope.
```suggestion
    let drop_fn = &(*lifetime).drop;
    drop_fn(owned)
```

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+}
+
+unsafe fn owned_drop_impl(owned: *mut ()) {
+    let lifetime = owned.cast::<OwnedLifetime>();
+    let ref_cnt = &(*lifetime).ref_cnt;
+
+    let old_cnt = ref_cnt.fetch_sub(1, Ordering::Acquire);
+    if old_cnt != 1 {
+        return;
+    }
```
Darksonn, line 1149: The subtraction must be a release. The acquire operation can happen afterwards.
```suggestion
    let old_cnt = ref_cnt.fetch_sub(1, Ordering::Release);
    if old_cnt != 1 {
        return;
    }
    ref_cnt.load(Ordering::Acquire);
```
This is also legal, but performs acquire fences more often than necessary:
```suggestion
    let old_cnt = ref_cnt.fetch_sub(1, Ordering::AcqRel);
    if old_cnt != 1 {
        return;
    }
```

src/bytes.rs
```diff
@@ -1012,13 +1081,99 @@ unsafe fn static_drop(_: &mut AtomicPtr<()>, _: *const u8, _: usize) {
...
+
+    let drop = &(*lifetime).drop;
+    drop(owned)
+}
+
+unsafe fn owned_drop(data: &mut AtomicPtr<()>, _ptr: *const u8, _len: usize) {
+    let owned = data.load(Ordering::Acquire);
```
Darksonn, line 1156:
```suggestion
    let owned = data.load(Ordering::Relaxed);
```

tests/test_bytes.rs
```diff
@@ -1479,3 +1482,139 @@ fn split_to_empty_addr_mut() {
...
+        if self.panic_as_ref {
+            panic!("test-triggered panic in `AsRef<[u8]> for OwnedTester`");
+        }
+        self.buf.as_slice()
+    }
+}
+
+impl<const L: usize> Drop for OwnedTester<L> {
+    fn drop(&mut self) {
+        self.drop_count.fetch_add(1, Ordering::AcqRel);
```
Darksonn, line 1514: These tests are single-threaded, so the orderings don't matter. :woman_shrugging:

src/bytes.rs
```diff
@@ -200,6 +202,69 @@ impl Bytes {
         }
     }
 
+    /// Create [Bytes] with a buffer whose lifetime is controlled
+    /// via an explicit owner.
+    ///
+    /// A common use case is to zero-copy construct from mapped memory.
+    ///
+    /// ```ignore
```
Darksonn, line 210: Instead of marking this test `ignore`, you can add a doc-hidden module called `memmap2` to simulate that the crate exists when compiling the example.

amunra:
I shall address the latest feedback tomorrow. I would appreciate clarity around what the behaviour of `is_unique` should be for `from_owner`.
If it is always to return `false` (the most pragmatic approach I think) I can update the API docs.

src/bytes.rs
amunra: I couldn't quite get that to work, as the `Mmap` type would otherwise be found via `use bytes::memmap2::Mmap` and not `memmap2::Mmap`, so I inlined the mock module instead.

tests/test_bytes.rs
amunra: True, but having these orderings littered everywhere isn't particularly readable, so I did some refactoring to tidy up into a `SharedAtomicCounter` struct.

src/bytes.rs
amunra: Sounds sensible, let me know and I can remove `cheap_into_mut`.

amunra: done

Darksonn (approved):
Thanks.

amunra:
Thank you!

paolobarbolini:
> Is there a plan to release a new version with this feature soon?

It makes sense to make a new release at this point. If you'd like feel free to send a release PR.

PR #748, opened 2024-11-27
Prepare bytes v1.9.0

kylebarron:
Ref https://github.com/tokio-rs/bytes/pull/742#issuecomment-2491268005.

A new release would be nice to take advantage of `Bytes::from_owner`


Darksonn (approved):
Thanks.

PR #751, opened 2024-12-22
Support non-atomic targets by using portable-atomic

tjoslin:
squashed commits from upstream #467 by @taiki-e and rebased with updated deps and resolved conflicts

I'm not sure this should be merged as-is because the original author would lose credit but given the comments on #467 I figured this was a useful exercise to see what else would be necessary to add before a merge could occur.


ia0:
> the original author would lose credit

In the worst case, you can always add `Co-authored-by: Taiki Endo <email redacted>` at the end of the commit(s) that will end up merged (copying the email from the commits of #467).

PR #752, opened 2024-12-28
The mark and rest methods have been added to Bytes and BytesMut

aohanhe:
The mark and rest methods have been added to Bytes and BytesMut. These methods enable quickly jumping back to the marked position when reading data, thus solving the problem that data decoding can only move forward and not backward. Due to time constraints, I haven't examined the original code carefully. Please point out any mistakes.


aohanhe:
I was a bit busy yesterday and forgot to add the unit tests.

PR #753, opened 2025-01-08
Try get methods for buf

### LINKED ISSUE #254: Request: Provide try_get_* methods

hpenne:
Adds fallible `try_get_` methods to Buf. The methods align with the existing `get_` methods.  All of them, so this is a lot of lines but most of it is documentation.  Note that I copied and adapted `buf_get_impl` to make `try_buf_get_impl`.  It is possible to avoid some code duplication by calling `buf_get_impl` instead, but that will checked the remaining bytes twice, and `buf_get_impl` is performance optimized so I chose not to do that.

This code borrows some from the existing PR for this feature by @caelunshun, but that is old with lots of conflicts, so I chose to reimplement it from the current master instead.

I would rater have used from_be_bytes etc. for the floats as that seems more portable for NaN values, but it seems that the supported rust version is perhaps one minor version too old for that.

Closes #254

### DIFF
--- a/src/buf/buf_impl.rs
+++ b/src/buf/buf_impl.rs
@@ -3,51 +3,19 @@ use crate::buf::{reader, Reader};
 use crate::buf::{take, Chain, Take};
 #[cfg(feature = "std")]
 use crate::{min_u64_usize, saturating_sub_usize_u64};
-use crate::{panic_advance, panic_does_not_fit};
+use crate::{panic_advance, panic_does_not_fit, TryGetError};
 
 #[cfg(feature = "std")]
 use std::io::IoSlice;
 
 use alloc::boxed::Box;
 
-/// Error type for the `try_get_` methods of [`Buf`].
-#[derive(Debug, PartialEq, Eq)]
-pub enum TryGetError {
-    /// Indicates that there were not enough remaining
-    /// bytes in the buffer while attempting
-    /// to get a value from a [`Buf`] with one
-    /// of the `try_get_` methods.
-    OutOfBytes {
-        /// The number of bytes necessary to get the value
-        requested: usize,
-
-        /// The number of bytes available in the buffer
-        available: usize,
-    },
-}
-
-impl core::fmt::Display for TryGetError {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
-        match self {
-            TryGetError::OutOfBytes {
-                requested,
-                available,
-            } => {
-                write!(f, "Not enough bytes remaining in buffer to read value (requested {} but only {} available)", requested, available)
-            }
-        }
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for TryGetError {}
-
 macro_rules! buf_try_get_impl {
     ($this:ident, $typ:tt::$conv:tt) => {{
         const SIZE: usize = core::mem::size_of::<$typ>();
 
         if $this.remaining() < SIZE {
-            return Err(TryGetError::OutOfBytes {
+            return Err(TryGetError {
                 requested: SIZE,
                 available: $this.remaining(),
             });
@@ -103,23 +71,17 @@ macro_rules! buf_try_get_impl {
 
 macro_rules! buf_get_impl {
     ($this:ident, $typ:tt::$conv:tt) => {{
-        return unwrap_advance((|| buf_try_get_impl!($this, $typ::$conv))());
+        return (|| buf_try_get_impl!($this, $typ::$conv))()
+            .unwrap_or_else(|error| panic_advance(&error));
     }};
     (le => $this:ident, $typ:tt, $len_to_read:expr) => {{
-        return unwrap_advance((|| buf_try_get_impl!(le => $this, $typ, $len_to_read))());
+        return (|| buf_try_get_impl!(le => $this, $typ, $len_to_read))()
+            .unwrap_or_else(|error| panic_advance(&error));
     }};
     (be => $this:ident, $typ:tt, $len_to_read:expr) => {{
-        return unwrap_advance((|| buf_try_get_impl!(be => $this, $typ, $len_to_read))());
+        return (|| buf_try_get_impl!(be => $this, $typ, $len_to_read))()
+            .unwrap_or_else(|error| panic_advance(&error));
     }};
-}
-
-fn unwrap_advance<T>(value: Result<T, TryGetError>) -> T {
-    value.unwrap_or_else(
-        |TryGetError::OutOfBytes {
-             requested,
-             available,
-         }| panic_advance(requested, available),
-    )
 }
 
 // https://en.wikipedia.org/wiki/Sign_extension
@@ -330,7 +292,8 @@ pub trait Buf {
     ///
     /// This function panics if `self.remaining() < dst.len()`.
     fn copy_to_slice(&mut self, dst: &mut [u8]) {
-        unwrap_advance(self.try_copy_to_slice(dst));
+        self.try_copy_to_slice(dst)
+            .unwrap_or_else(|error| panic_advance(&error));
     }
 
     /// Gets an unsigned 8 bit integer from `self`.
@@ -351,7 +314,10 @@ pub trait Buf {
     /// This function panics if there is no more remaining data in `self`.
     fn get_u8(&mut self) -> u8 {
         if self.remaining() < 1 {
-            panic_advance(1, 0);
+            panic_advance(&TryGetError {
+                requested: 1,
+                available: 0,
+            })
         }
         let ret = self.chunk()[0];
         self.advance(1);
@@ -376,7 +342,10 @@ pub trait Buf {
     /// This function panics if there is no more remaining data in `self`.
     fn get_i8(&mut self) -> i8 {
         if self.remaining() < 1 {
-            panic_advance(1, 0);
+            panic_advance(&TryGetError {
+                requested: 1,
+                available: 0,
+            });
         }
         let ret = self.chunk()[0] as i8;
         self.advance(1);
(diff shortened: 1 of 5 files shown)


src/buf/buf_impl.rs
```diff
@@ -10,6 +10,30 @@ use std::io::IoSlice;
...
 
+/// Error type for the `try_get_` methods of [`Buf`].
+#[derive(Debug, PartialEq, Eq)]
+pub enum TryGetError {
+    /// Indicates that there were not enough remaining
+    /// bytes in the buffer while attempting
+    /// to get a value from a [`Buf`] with one
+    /// of the `try_get_` methods.
+    OutOfBytes,
+}
```
Darksonn, line 21: We can make this into a struct with two integers for the number of requested and actually available byte counts.

src/buf/buf_impl.rs
```diff
@@ -1113,6 +1195,1205 @@ pub trait Buf {
...
+            let src = self.chunk();
+            let cnt = usize::min(src.len(), dst.len());
+
+            dst[..cnt].copy_from_slice(&src[..cnt]);
+            dst = &mut dst[cnt..];
+
+            self.advance(cnt);
+        }
+        Ok(())
+    }
```
Darksonn, line 1243: Would it make sense to implement the non-try methods by calling the try method?

hpenne: I think that will add an extra check of the return value from the try method.  Doing it the other way around means checking the remaining size twice instead of once.  If this was my day job I would be inclined to do one of them to reduce code duplication, but your code looked like you had put in an solid effort to optimise it so I wasn't sure if that would be acceptable.

src/buf/buf_impl.rs
hpenne: Will do.

src/buf/buf_impl.rs
Darksonn: I imagine we could implement `copy_to_slice` as `self.try_copy_to_slice().unwrap()` without introducing extra cost?

hpenne: Absolutely. Would you prefer panic_advance to unwrap?

My other concern is that the two macros are quite similar. If I make the get-macros use the try-macro then a good optimiser might even make that cost free. Is that an OK improvement?

hpenne: There might be a small extra cost for copy_to_slice (the unwrap and the creation of the Result return value), but I imagine it depends a lot on how smart the compiler is.  It might get removed during optimisation if the call to try_copy_to_slice is inlined by the compiler.
For the macros it will all be inside the same function, so I imagine the optimiser is more likely to remove any overhead.

hpenne: I've had a go at removing code duplication. I don't think there should be much of a performance hit in optimised builds (if any), but I have not made any measurements.

src/buf/buf_impl.rs
```diff
@@ -10,12 +10,48 @@ use std::io::IoSlice;
...
+    OutOfBytes {
+        /// The number of bytes necessary to get the value
+        requested: usize,
+
+        /// The number of bytes available in the buffer
+        available: usize,
+    },
+}
+
+#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
hpenne: Thanks.  I missed that part.  The existing 128 bit get methods seems to be missing from there as well.  Can I add those while I'm at it, or is the omission deliberate?

src/buf/buf_impl.rs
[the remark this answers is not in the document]
hpenne: Thanks. I missed that part. The existing 128 bit get methods seem to be missing from there. Can I add those as well, or is the omission deliberate somehow?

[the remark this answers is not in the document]
Darksonn: Are you adding new methods that are not just `try_*` variants of existing methods? Which ones?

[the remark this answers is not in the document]
hpenne: @Darksonn What I suggested was to add to the `deref_forward_buf` macro support for the 128 bit get methods (like `get_i128`, `get_i128_le` and so forth), in addition to the `try_get` methods. The 128 bit get methods seemed to be missing, and it is easy work to add them while I add the `try_get` methods.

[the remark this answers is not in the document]
Darksonn: Ah, ok, so you're just adding them to `deref_forward_buf`, not adding new methods. That's okay with me.

src/buf/buf_impl.rs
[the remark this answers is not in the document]
hpenne: Can that be made to work? Isn't `Display` in `std`?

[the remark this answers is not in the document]
paolobarbolini: `Display` is in `core`. Like everything else `std` re-exports it.
https://doc.rust-lang.org/nightly/core/fmt/trait.Display.html

src/buf/buf_impl.rs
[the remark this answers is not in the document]
hpenne: Added the new methods, plus the missing existing get methods.

src/buf/buf_impl.rs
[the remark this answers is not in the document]
hpenne: Embarrassing. Thank you for your patience. Writing "no_std" code is new to me.
Pushed the change.

hpenne:
@Darksonn @paolobarbolini I think I have made the necessary changes to the comments I have received so far. I have pushed the "Resove" button on each to indicate this (I hope this was the correct thing to do in this repo, it seems to vary). I hope you have the time next week to review my changes. Also, the CI needs to be started on the latest changes.

Darksonn:
Thanks, clicking resolve is fine. :+1:

src/buf/buf_impl.rs
```diff
@@ -10,12 +10,47 @@ use std::io::IoSlice;
...
 
-macro_rules! buf_get_impl {
+/// Error type for the `try_get_` methods of [`Buf`].
+#[derive(Debug, PartialEq, Eq)]
+pub enum TryGetError {
+    /// Indicates that there were not enough remaining
+    /// bytes in the buffer while attempting
+    /// to get a value from a [`Buf`] with one
+    /// of the `try_get_` methods.
+    OutOfBytes {
```
Darksonn, line 20: I think this can just be a struct.

src/buf/buf_impl.rs
```diff
@@ -10,12 +10,47 @@ use std::io::IoSlice;
 
 use alloc::boxed::Box;
 
-macro_rules! buf_get_impl {
+/// Error type for the `try_get_` methods of [`Buf`].
+#[derive(Debug, PartialEq, Eq)]
+pub enum TryGetError {
```
Darksonn, line 15: We may want to implement `From<TryGetError>` for `std::io::Error`. This will make it work better with the `?` operator.

src/buf/buf_impl.rs
```diff
@@ -61,9 +96,30 @@ macro_rules! buf_get_impl {
...
+        return unwrap_advance((|| buf_try_get_impl!(be => $this, $typ, $len_to_read))());
+    }};
+}
+
+fn unwrap_advance<T>(value: Result<T, TryGetError>) -> T {
+    value.unwrap_or_else(
+        |TryGetError::OutOfBytes {
+             requested,
+             available,
+         }| panic_advance(requested, available),
```
Darksonn, line 121: This function could take the error type as argument.

src/buf/buf_impl.rs
hpenne: True. My experience is that even so, it often pays off to have an outer enum because it allows more error types to be added in the future with less API breakage.  If we make it a struct then you may need a new major version to add another error type here. If you prefer a struct then I will of course change it.

Darksonn: Your current enum cannot have additional error types added. You need to mark it `#[non_exhaustive]` if that was your intent.

hpenne: Of course. My habit here is from in-house projects. I suppose there is no good way to add additional error types here in the future that doesn't break something, so a struct sounds like a good solution.

src/buf/buf_impl.rs
hpenne: Will do.

src/buf/buf_impl.rs
hpenne: I'm afraid I do not understand this one. By "This function" do you refer to `panic_advance` or `unwrap_advance`, and what should this argument be used for? It sounds like the argument is a means to an end, but I do not see what this end is. Sorry for being slow on the uptake...

Darksonn: I mean that `panic_advance` could take an `TryGetError` argument instead of two integers. That way, you can't accidentally mix up the order of `requested` and `available` when calling it.

hpenne: So a refactoring of panic_advance. That is a good idea, especially when TryGetError is changed. It makes argument mixups impossible and makes the code inside `unwrap_advance` so simple that it is perhaps not necessary to have a separate function for this any more. Thanks.

hpenne: Made the change. Ended up removing unwrap_advance, as panic_advance now seems sufficient. I also moved TryGetError to lib.rs to have it on the same level as panic_advance. It seemed odd to have panic_advance have a dependency on an error defined lower down. I had some doubts if that was the right thing to to, so let me know if this was a mistake and I'll move it back.

Darksonn (approved):
Thank you.

PR #763, opened 2025-01-25
ci: test AArch64 Linux and run tests instead of build in cross job

taiki-e:
## Motivation

Currently, "cross" job in our CI is build-only and doesn't run tests:

https://github.com/tokio-rs/bytes/blob/3ab876fee66393b064c2adac83d28fca46186075/.github/workflows/ci.yml#L121-L129

And AArch64 Linux which is tier1 in rustc is not tested.

https://github.com/tokio-rs/bytes/blob/3ab876fee66393b064c2adac83d28fca46186075/.github/workflows/ci.yml#L106-L113

## Solution

This PR adds `aarc64-unknown-linux-gnu` (ubuntu-22.04-arm) to matrix of tests on stable ("stable" job), and update "cross" job to run tests by using setup-cross-toolchain-action which is used in tokio. (see also https://github.com/tokio-rs/tokio/pull/7123 for AArch64/Armv7hf)

As for WASM testing, use `wasm32-wasip1` instead of `wasm32-unknown-unknown` because `wasm32-wasip1` can test with `cargo test` command. (Testing `wasm32-unknown-unknown` usually needs `wasm-pack` or other commands.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
+          - ubuntu-22.04-arm # TODO: update to 24.04 when https://github.com/rust-lang/rust/issues/135867 solved
           - macos-latest
           - windows-latest
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks.

PR #772, opened 2025-03-04
add owned_vec_into test

venomousmoog:


venomousmoog:
oops - didn't mean to create this with just a failing test

PR #773, opened 2025-03-04
Memory leak in owned_to_vec

venomousmoog:
The following sequence:

```rust
    let b1 = Bytes::from_owner(owner);
    let v1: Vec<u8> = b1.into();
```

Would fail to drop `owner` (the expectation for the vtable *_to_vec is that it takes ownership of the instance, but owned_to_vec was not).  Since owned_to_mut depended on owned_to_vec and was correctly dropping, the fix was to move the drop logic from owned_to_mut to owned_to_vec.

This change adds an appropriate test for ::into() for Bytes::from_owner.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -1149,9 +1149,7 @@ unsafe fn owned_to_vec(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> Vec<
 }
 
 unsafe fn owned_to_mut(data: &AtomicPtr<()>, ptr: *const u8, len: usize) -> BytesMut {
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thank you.

scottlamb:
> the expectation for the vtable *_to_vec is that it takes ownership of the instance, but owned_to_vec was not

Should the `to_vec` and `to_mut` vtable entries be called `into_vec` and `into_mut`, respectively, to make this misunderstanding less likely?

I also wonder if the `owned_to_vec` test in `test_bytes.rs` is accomplishing anything, given that it's using `slice::to_vec` rather than a `Bytes` method. I suspect it was meant to exercise the behavior the new `owner_into_vec` test does.

PR #781, opened 2025-03-24
Add get_array method to Buf

lperlaki:
This would increase the MSRV to 1.51 since it uses const generics!

I am not sure on the MSRV policy of bytes, so if this is a problem we can just close this PR

### Added Methods

- `Buf::try_get_array`
    is a copy of the old implementation of `buf_try_get_impl`
- `Buf::get_array` simply calls `try_get_array` and `panic_advance` on error

the second commit  changes the implementation of the `get_x` methods to use `try_get_array`

this would also enable `Buf` implementors to optimise (like #622) the `get_x` methods by overriding `try_get_array`


lperlaki:
as related work there is also a std lib proposal to add a `std::io::Read::read_array` https://github.com/rust-lang/rust/issues/148848

PR #806, opened 2025-11-21
Fix `get_int` if `nbytes` is zero

### LINKED ISSUE #798: Panic in `get_int` if nbytes is zero
`get_int` panics if nbytes is zero due to attempt to right shift by 64 when doing the sign extension.

```
    let mut buf = &b"hello world"[..];
    let ret1 = buf.get_int(0);
```

Not sure if this is the intended behaviour  that isn't documented. Doing the same with `get_uint` returns zero which might be preferrable.

paolobarbolini:
Fixes #798


Darksonn (approved):
Thanks.

Issue #805, opened 2025-11-19
std::io::Read into BytesMut efficiency


alexheretic:
I worked on [_tungstenite_ code](https://github.com/snapview/tungstenite-rs/blob/master/src/protocol/frame/mod.rs#L236) which uses `std::io::Read::read` into a `BytesMut` buffer.

One issue with this is to provide the full capacity to the `Read` impl `BytesMut::resize` is called on each read to zero fill the capacity so that it is safely usable. 

This isn't optimal as it may require re-writing zeros across say a larger buffer on small reads. I was wondering if there is prior art on solving this issue?

Potential solutions:
* Use `set_len`, but this is not sound as `Read` impls _could_ read the buffer before writing to it. ❌ 
* Write a variant of `BytesMut` that always zeroes its capacity. It would then be sound to use set_len in this case. Does anything like this already exist?

alexheretic:
Thanks for both your responses. I've proposed a custom optimisation for this for [tungstenite](https://github.com/snapview/tungstenite-rs/pull/524) & look forward to using https://github.com/rust-lang/rust/issues/78485 when available.

PR #808, opened 2025-11-23
Document that `BytesMut::{reserve,try_reserve}` doesn't preserve unused capacity

paolobarbolini:
This documents that `BytesMut` may not preserve the memory outside of `0..len` after calling `BytesMut::reserve`. It may for example discard all of it, and replace it with uninit memory in the [fallback case](https://github.com/tokio-rs/bytes/blob/acd1e0ffb8f076225759b8005d04f65ef77cccca/src/bytes_mut.rs#L767-L780), or even shift it causing a different region of the `BytesMut` to appear when going through the [reclaim optimization](https://github.com/tokio-rs/bytes/blob/acd1e0ffb8f076225759b8005d04f65ef77cccca/src/bytes_mut.rs#L685-L711).

This was encountered in https://github.com/snapview/tungstenite-rs/pull/524#discussion_r2554375051 after #805 was closed, which could have lead to unsound code being merged into `tungstenite`.

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -551,6 +551,8 @@ impl BytesMut {
     /// and the original buffer is large enough to fit the requested additional
     /// capacity, then reallocations will never happen.
     ///
+    /// This method does not preserve data stored in the unused capacity.
+    ///
     /// # Examples
     ///
     /// In the following example, a new buffer is allocated.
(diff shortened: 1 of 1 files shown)


src/bytes_mut.rs
```diff
@@ -551,6 +551,9 @@ impl BytesMut {
     /// and the original buffer is large enough to fit the requested additional
     /// capacity, then reallocations will never happen.
     ///
+    /// Once this method has been called, the contents of memory outside of
+    /// `0..len` are undefined and may be uninitialized.
```
Darksonn: The word undefined is somewhat ambiguous in this context. Let's avoid that. How about this?
```suggestion
    /// This method does not preserve data stored in the unused capacity.
```

PR #813, opened 2025-12-22
style: address some clippy warnings

danieleades:
addresses some clippy warnings (including some pedantic ones)

i see there's a commented out clippy linting job- does that mean there's an ambition to sort out the warnings and enable linting?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 1.11.1 (February 3rd, 2026)
+
+- Fix integer overflow in `BytesMut::reserve`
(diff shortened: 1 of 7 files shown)


martin-g:
> i see there's a commented out clippy linting job- does that mean there's an ambition to sort out the warnings and enable linting?

This would be nice!

danieleades:
> > i see there's a commented out clippy linting job- does that mean there's an ambition to sort out the warnings and enable linting?
> 
> This would be nice!

I can take a look at that in a future PR

PR #819, opened 2026-01-23
Implement `Bytes::unsplit` and `Bytes::try_unsplit`

### LINKED ISSUE #503: Consider an unsplit method for Bytes

thanhminhmr:
Closes #503, #287.

- [x] Implement `Bytes::unsplit` and `Bytes::try_unsplit`
- [x] Implement tests for `Bytes::unsplit` and `Bytes::try_unsplit`

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -172,7 +172,7 @@ impl Bytes {
         Bytes {
             ptr: bytes.as_ptr(),
             len: bytes.len(),
-            data: AtomicPtr::new(ptr::null_mut()),
+            data: AtomicPtr::new(bytes.as_ptr() as *mut ()),
             vtable: &STATIC_VTABLE,
         }
     }
@@ -183,7 +183,7 @@ impl Bytes {
         Bytes {
             ptr: bytes.as_ptr(),
             len: bytes.len(),
-            data: AtomicPtr::new(ptr::null_mut()),
+            data: AtomicPtr::new(bytes.as_ptr() as *mut ()),
             vtable: &STATIC_VTABLE,
         }
     }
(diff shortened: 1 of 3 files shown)


Darksonn:
I'm sorry but this PR raises a lot of forwards compatibility concerns for me. I'm concerned that it may limit our ability to add other features in the future. See #287 for prior discussion.

thanhminhmr:
To clarify how the current PR aligns with that contract:

The implementation only attempts to merge when it can *prove* the two `Bytes` share the same backing data. Concretely, it compares the internal `data` pointer and only proceeds if they are equal; otherwise it conservatively fails.

For `Bytes::from_static`, I updated the (currently unused) `data` field to store the original address of the static slice. That address is preserved across splits, so `Bytes` originating from the same static allocation share the same `data`. This allows static `Bytes` to be safely merged when appropriate and fixes the issue you pointed out earlier. I verified this behavior with `miri`, and it reports no UB after the fix.

There is one edge case I have not handled yet: merging when `self` is empty and `other` is non-empty. The current logic will conservatively fail if their backing data differ. This could be optimized by dropping `self` and moving `other` into it without copying, but I’ve intentionally left it out for now to keep the initial implementation minimal and conservative.

scutuatua-crypto (approved):

PR #822, opened 2026-03-05
Adds `Buf` impl for `VecDeque<impl Buf>`

b01o:
This PR adds an implementation of the `Buf` trait for `VecDeque<impl Buf>`.

Sometimes we need to store multiple non-contiguous `Bytes` at runtime. Using `Chain` can be limiting because it doesn't provide a concrete type for a collection of buffers. Implementing `Buf` for `VecDeque<impl Buf>` gives a handy way to view these disjoint chunks as a single logical buffer.

#### Update: 
- `Buf` is implemented for `VecDeque<impl Buf>` instead of `VecDeque<Bytes>`. 
- implements `chunks_vectored` for `VecDeque<impl Buf>`
- only return empty chunks if `remaining()` is 0
- adds a test for `chunks_vectored()`


src/buf/vec_deque.rs
```diff
@@ -38,3 +38,33 @@ impl Buf for VecDeque<u8> {
         self.drain(..cnt);
     }
 }
+
+impl Buf for VecDeque<crate::Bytes> {
```
Darksonn: If we're going to do that, I think it would be nicer to expand the impl for other buffer types.
```suggestion
impl<T: Buf> Buf for VecDeque<T> {
```

b01o: Nice touch!

src/buf/vec_deque.rs
```diff
@@ -38,3 +38,33 @@ impl Buf for VecDeque<u8> {
         self.drain(..cnt);
     }
 }
+
+impl<T: Buf> Buf for VecDeque<T> {
+    fn remaining(&self) -> usize {
+        self.iter().map(|b| b.remaining()).sum()
+    }
+
+    fn chunk(&self) -> &[u8] {
```
Darksonn, line 47: Please implement `chunks_vectored` too.

b01o: Added.

src/buf/vec_deque.rs
```diff
@@ -38,3 +38,45 @@ impl Buf for VecDeque<u8> {
...
+            n += buf.chunks_vectored(&mut dst[n..]);
+        }
+        n
+    }
+
+    fn advance(&mut self, mut cnt: usize) {
+        while cnt > 0 {
+            let Some(b) = self.front_mut() else {
+                panic!("advance called with cnt > remaining");
+            };
```
paolobarbolini: We can't use the let-else syntax with MSRV v1.57

b01o: Fixed.

src/buf/vec_deque.rs
```diff
@@ -38,3 +38,43 @@ impl Buf for VecDeque<u8> {
...
+        self.iter().map(|b| b.remaining()).sum()
+    }
+
+    fn chunk(&self) -> &[u8] {
+        if let Some(b) = self.front() {
+            b.chunk()
+        } else {
+            &[]
+        }
+    }
```
paolobarbolini, line 52: Now that I think of it more, this violates the implementer notes on `chunk()`. Since we accept any `VecDeque`, there could be empty elements within it, something that common `BufList` implementations prevent [^1] [^2]

> `chunk()` should return an empty slice **if and only if** `remaining()` returns 0

[^1]: https://docs.rs/crate/hyper/1.8.1/source/src/common/buf.rs#17-21
[^2]: https://docs.rs/crate/watermelon-proto/0.1.8/source/src/util/buf_list.rs#22-27

b01o: I see what you mean, let me try to make it follow the contract.

src/buf/vec_deque.rs
```diff
@@ -38,3 +38,48 @@ impl Buf for VecDeque<u8> {
...
+
+    fn chunk(&self) -> &[u8] {
+        if !self.has_remaining() {
+            return &[];
+        }
+
+        self.iter()
+            .find(|b| b.has_remaining())
+            .map(|b| b.chunk())
+            .expect("chunk called with no remaining bytes")
```
paolobarbolini: It should be more efficient to scan the items once (just `self.iter()`) instead of twice (`has_remaining` + `self.iter()`)

```suggestion
        self.iter()
            .find(|b| b.has_remaining())
            .map(|b| b.chunk())
            .unwrap_or_default()
```

src/buf/vec_deque.rs
```diff
@@ -38,3 +38,44 @@ impl Buf for VecDeque<u8> {
...
+    }
+
+    #[cfg(feature = "std")]
+    fn chunks_vectored<'a>(&'a self, dst: &mut [io::IoSlice<'a>]) -> usize {
+        let mut n = 0;
+        for buf in self {
+            if n >= dst.len() {
+                break;
+            }
+            n += buf.chunks_vectored(&mut dst[n..]);
```
Darksonn, line 63: This loop must exit if the chunks written by `buf` have a combined length smaller than `buf.remaining()`, because you must read everything from the first buffer before you start reading from the second buffer. Please add a test that would have caught this.

src/buf/vec_deque.rs
Darksonn: If we need a bugfix to existing code, it should be a separate PR.

src/buf/vec_deque.rs
b01o: The access issue should be fixed now. Let me know if I missed anything.

src/buf/vec_deque.rs
```diff
@@ -38,3 +38,113 @@ impl Buf for VecDeque<u8> {
...
+            } else {
+                cnt -= rem;
+                self.pop_front();
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
Darksonn: Tests go in `tests/` directory outside of `src/` unless they *have* to be here.

b01o: Now under `tests/test_buf.rs`.

Issue #823, opened 2026-03-10
Suggestion(V2): change `Buf::chunk() -> &[u8]` to `Buf::nth_chunk(usize) -> &[u8]`


b01o:
I wonder if is it possible to change `chunk()` to return any position of it's underlying slice instead. Currently we always returns the first chunk in a `Buf`, which makes it hard to visit a element not within the first chunk. To support access on later chunks, I propose that we could alter the `Buf::chunk() -> &[u8]` to `Buf::nth_chunk(usize) -> &[u8]`.

``` rust
trait bytes_v1::Buf {
    fn chunk(&self) -> &[u8];
    //...
}

trait bytes_v2::Buf {
    /// This function should never panic. `nth_chunk(0)` (i.e. `chunk()`)
    /// should return an empty slice **if and only if** `remaining()` returns 0.
    /// In other words, `chunk()` returning an empty slice implies that
    /// `remaining()` will return 0 and `remaining()` returning 0 implies that
    /// `chunk()` will return an empty slice.
    ///
    /// If `nth_chunk(n)` returns an empty slice, then `nth_chunk(m)` must also
    /// return an empty slice for all `m >= n`.
    fn nth_chunk(&self, nth: usize) -> &[u8];
    fn chunk(&self) -> &[u8] { self.nth_chunk(0) }
    // ...
}
```

#### Motivation: 

* easy to implement peek_* methods access later bytes without advance the buffer first, useful for many codec parsing.

```rust
trait bytes_v2::Buf {
    fn peek(&self, &mut [u8]) -> usize;
    fn peek_exact(&self, &mut [u8]) -> Result<(), TryGetError>;
    fn peek_u32(&self) -> Result<u32, TryGetError>;
    // ...
}
```

* should be enable to make `bytes_v1` depends on `bytes_v2` with this. I don't know how exactly the "semver hack" works, but I assume it's doable as we can simply forward all `bytes_v1::chunk()` to `bytes_v2::chunk()` ?
* enable us to optimize data structure more aggressively, since we now have the whole picture what's upfront.

#### Drawbacks: 
* changing interface means we need to re-implement all `chunk()` in current data structures, including `Chain`, `Take` and possibly more.
* more complex/error-prone when implementing new data structures, since with `bytes_v1` we only consider the first chunk.

Issue #382, opened 2020-03-28
API to peek at integers of buffer


mandreyel:
This is a feature request for the `Buf` trait to provide an API to peek at an integer of the buffer without advancing the underlying cursor. E.g. there is [`get_u32`](https://docs.rs/bytes/0.5.4/bytes/buf/trait.Buf.html#method.get_u32) and it would be really useful to have the equivalent but non cursor advancing `peek_u32` that returns the next `u32` of buffer.

If this feature already exists, my apologies, but I seem to have missed it. If not, is there any reason it is not included?

carllerche:
Peeking is a bit harder. `Buf` is basically a byte optimized iterator. In order to provide peaking of multiple bytes, you need to be able to rewind. Rewinding is not trivial to do for some implementations (single linked list, for example).

There are a few options. First, you can take `T: Buf + Clone`. In which case, this implies that you can clone the buffer to "peek".

Another alternative is something I proposed in https://github.com/tokio-rs/bytes/issues/300, but this would work only for "bufs" that are in continuous memory.

mandreyel:
@carllerche Those are fair points.

Maybe this could be a separate trait implemented only by buffers that abstract over contiguous memory, such as `BytesMut` (at least as far as I understand it is contiguous, by inspecting its [source](https://docs.rs/bytes/0.5.4/src/bytes/bytes_mut.rs.html#49-54)? Maybe the trait could be called `Peek` and offer the `peek_<integer>` methods. What do you think? Have I missed something?

---

This is a bit unrelated, but I wanted to give you an idea since my use case involves `tokio_codec` and as far as I know, you are its author.

My use case of this in implementing the `Decoder` trait is to decode in a TCP based binary protocol, where messages are prefixed with a fixed size length header, whether the buffer has the full message, and only proceeding to decode (i.e. consume the bytes in the buffer) if it does. Now, this only works if the message length prefix is kept in the buffer until the whole message can be extracted (thus the need for peeking).

Currently I work around this limitation with the following pattern, exploiting the fact that `Buf` is implemented for byte slices:
```rust
let mut buf: BytesMut = ...;

// ...

let mut tmp_buf = buf.bytes();
let payload_len = tmp_buf.get_u32() as usize;

// check that we got the full payload in the buffer (NOTE: we need
// to add the message length prefix's byte count to payload_len
// since the buffer cursor was not advanced and thus we need to
// consider the prefix too)
if buf.remaining() >= 4 + payload_len {
[...]

PR #825, opened 2026-03-29
Introduce Cursor and CursorMut traits

mhebant:
Has many others before, I've been looking for a way to read or write ahead, or seek into a `Buf` or a `BufMut` (see #823, #658, #382 to name a few)

While this PR might not solve everyone's problems, I has the benefit of not braking the API (it only adds 2 new traits) and simple (by relying as much as possible on the existing `Buf` and `BufMut`).

The simple solution to read ahead in a `Buf` I've seen proposed (like here https://github.com/tokio-rs/bytes/issues/382#issuecomment-605513824) is to clone the `Buf`. But making `T: Buf + Clone` your interface wont work in practice: it's fine if your `Buf` is a `&[u8]` but not if it's a `Vec<u8>`.

So the Idea is to add a trait (named Cursor for now) whit an associated type that is a cheap version of our `Buf`.
For example the Cursor of a `Vec<u8>` is a `&[u8]`, for a `Chain<Vec<u8>, Vec<u8>>` it's a `Chain<&[u8], &[u8]>` etc...

The Cursor trait allows you to get a cheap copy of you `Buf` that you can use to read ahead without mutating the original `Buf`.

This should also work for writing ahead to a `BufMut` before possibly committing what have been written by calling advance.

This PR is incomplete, lacks some implementation and tests, but I'm looking for feedback before finishing it.

Also I understand adding new features is complicated because any it's impossible to make breaking change. It should be possible to put this in it's own create depending on bytes. But again looking for some feedback.

Thanks !

### DIFF
--- a/src/buf/chain.rs
+++ b/src/buf/chain.rs
@@ -1,5 +1,5 @@
 use crate::buf::{IntoIter, UninitSlice};
-use crate::{Buf, BufMut};
+use crate::{Buf, BufMut, Cursor, CursorMut};
 
 #[cfg(feature = "std")]
 use std::io::IoSlice;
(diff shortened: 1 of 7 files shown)


Darksonn:
I've spent some time thinking about this. Unfortunately I think the desire to keep bytes simple outweighs this feature request, so I'm going to close it.

PR #827, opened 2026-05-18
chore: improve bytes maintenance path

vip892766gma:
Summary:
- Audit the `BytesMut`/`Bytes` unsafe blocks for missing or imprecise SAFETY comments and add a small targeted test (or doc-test) covering a boundary condition such as `split_to`/`split_off` at index 0 and at `len()`, and verify capacity/refcount invariants — a small code-level patch (test + a SAFETY comment clarification) with no API change.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -489,6 +489,8 @@ impl Bytes {
 
         self.len = at;
 
+        // SAFETY: `at` has been asserted to be <= `self.len()`, and the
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
Thanks!

PR #829, opened 2026-06-15
Specialize `BytesMut::extend<u8>` when its Vec-backed

AdamGS:
This PR adds a specialized code path to extending `BytesMut` when its backed by a `Vec`. There are no external changes, only a private function and type with some tests.

When benchmarking on my laptop (M3 Max MacBook Pro), the change to `put_slice_vec_extend` is from 480 MB/s to over 9000MB/s. Happy to bench on some Intel hardware if the data is interesting.

### DIFF
--- a/src/bytes_mut.rs
+++ b/src/bytes_mut.rs
@@ -1197,6 +1197,26 @@ impl BytesMut {
             slice::from_raw_parts_mut(ptr.cast(), len)
         }
     }
+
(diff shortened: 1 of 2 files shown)


AdamGS:
Messed up an unsafe block, now fixed

AdamGS:
I've made another pass at this PR to simplify it. Its mush smaller and simpler now while still having the same performance.

PR #837, opened 2026-07-05
Handle unwinding from Box::new

### LINKED ISSUE #836: Exception safety violations in `BytesMut`, leading to Double-Free
Hello, we are security researchers targeting Rust security. By running our tool in this repository, we found a soundness issue. **This report is written by 100% human.** We promise that all you read will never be generated by LLM.

We want to emphasize that, the issue we want to report is very corner case, and may not be considered as a soundness issue (because it is related to OOM triggered by `Box::new`, which may not be taken into consideration). If you don't think this is worth fixing, we apologize for the valuable time spent.

The issue happens in `BytesMut::promote_to_shared`:

https://github.com/tokio-rs/bytes/blob/91402cee605b21bf530ac07ede463599182b5d32/src/bytes_mut.rs#L1097-L1128

In line 1115, a `Box` to `Shared` is created, and then the shared pointer is stored as `self.data` in normal path. However, if `Box::new` panics due to OOM (can be triggered easily via a customized global allocator, and need to set oom behavior to panic), the `shared` structure is however already created with `shared.vec` **owns** the data pointed by `self.ptr`. As a result, when the panic happens, both `shared` and `self` (which is a `BytesMut`) will be dropped, so the actual memory will be double freed.

The similar issue also happens for `Bytes::shallow_clone_vec`:

https://github.com/tokio-rs/bytes/blob/91402cee605b21bf530ac07ede463599182b5d32/src/bytes.rs#L1456-L1482

Darksonn:
This fixes a few cases where a panic from `Box::new` has incorrect behavior because it runs the destructor of the argument passed to `new`. They are fixed by ensuring that the allocation is performed before creating the value to be stored in the box.

Note that the default configuration is to abort on allocation failure, but it's possible to configure the allocator so that OOM results in panic instead of abort.

Fixes: #836

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -946,17 +946,16 @@ impl From<&'static str> for Bytes {
 
 impl From<Vec<u8>> for Bytes {
     fn from(vec: Vec<u8>) -> Bytes {
+        // Avoid an extra allocation if possible.
+        if vec.len() == vec.capacity() {
+            return Bytes::from(vec.into_boxed_slice());
+        }
+
         let shared = Box::new(MaybeUninit::<Shared>::uninit());
         let mut vec = ManuallyDrop::new(vec);
         let ptr = vec.as_mut_ptr();
         let len = vec.len();
         let cap = vec.capacity();
-
-        // Avoid an extra allocation if possible.
-        if len == cap {
-            let vec = ManuallyDrop::into_inner(vec);
-            return Bytes::from(vec.into_boxed_slice());
-        }
 
         let shared = Shared::init_to_raw(
             shared,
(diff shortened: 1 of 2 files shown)


src/bytes.rs
```diff
@@ -946,6 +946,7 @@ impl From<&'static str> for Bytes {
 
 impl From<Vec<u8>> for Bytes {
     fn from(vec: Vec<u8>) -> Bytes {
+        let shared = Box::new(MaybeUninit::<Shared>::uninit());
```
conradludgate, line 954: This seems to go against the line
> // Avoid an extra allocation if possible

Darksonn: Good point. I wanted to move it before the ManuallyDrop line. Fixed.

seanmonstar (approved):

PR #839, opened 2026-07-08
fix: try_get_int/try_get_int_le do not sign-extend n-byte integers

teddytennant:
## What

`Buf::try_get_int` and `Buf::try_get_int_le` now sign-extend the value they read for `nbytes < 8`, matching the behavior of the panicking `Buf::get_int` / `Buf::get_int_le`.

## Why

The panicking getters sign-extend via `sign_extend(self.get_uint(nbytes), nbytes)`. Their fallible twins instead used `buf_try_get_impl!`, which copies the bytes into the low bytes of a zeroed `[0u8; 8]` and calls `i64::from_{be,le}_bytes`. For `nbytes < 8` the high bytes stay zero, so no sign extension happens and a negative value is returned as a wrong positive number.

For example, the 3-byte big-endian value `0xffffff` (which `get_int(3)` returns as `-1`) was returned by `try_get_int(3)` as `16777215`.

`try_get_int_ne` delegates to these two methods and inherited the bug; it is fixed transitively. No `Buf` implementation overrides these methods, so all implementors were affected.

The fix makes the fallible getters mirror their panicking counterparts:

```rust
fn try_get_int(&mut self, nbytes: usize) -> Result<i64, TryGetError> {
    Ok(sign_extend(self.try_get_uint(nbytes)?, nbytes))
}
fn try_get_int_le(&mut self, nbytes: usize) -> Result<i64, TryGetError> {
    Ok(sign_extend(self.try_get_uint_le(nbytes)?, nbytes))
}
```

Edge cases are preserved: `nbytes > 8` still panics via `try_get_uint`'s documented check, `nbytes == 8` is an identity shift, and insufficient data still returns the same `Err(TryGetError { requested, available })` as before (existing doctests unchanged).

## Testing

- Added `try_get_int_sign_extends` to `tests/test_buf.rs`, covering big/little/native-endian sign extension, a positive value, full width, and the insufficient-data error shape. It fails on `master` (`try_get_int(3)` returns `Ok(16777215)`) and passes with this change.
- `cargo test` passes all suites and doctests, including the existing `try_get_int`/`try_get_int_le` doctests.
- `cargo fmt --all --check` is clean.

### DIFF
--- a/src/buf/buf_impl.rs
+++ b/src/buf/buf_impl.rs
@@ -2084,7 +2084,7 @@ pub trait Buf {
     ///
     /// This function panics if `nbytes` is greater than 8.
     fn try_get_int(&mut self, nbytes: usize) -> Result<i64, TryGetError> {
-        buf_try_get_impl!(be => self, i64, nbytes);
+        Ok(sign_extend(self.try_get_uint(nbytes)?, nbytes))
     }
 
     /// Gets a signed n-byte integer from `self` in little-endian byte order.
@@ -2116,7 +2116,7 @@ pub trait Buf {
     ///
     /// This function panics if `nbytes` is greater than 8.
     fn try_get_int_le(&mut self, nbytes: usize) -> Result<i64, TryGetError> {
-        buf_try_get_impl!(le => self, i64, nbytes);
+        Ok(sign_extend(self.try_get_uint_le(nbytes)?, nbytes))
     }
 
     /// Gets a signed n-byte integer from `self` in native-endian byte order.
--- a/tests/test_buf.rs
+++ b/tests/test_buf.rs
@@ -459,3 +459,35 @@ fn copy_to_bytes_mut() {
     let bytes_mut2 = BytesMut::from(ret);
     assert_eq!(bytes_mut2.as_ptr(), ptr);
 }
+
+#[test]
+fn try_get_int_sign_extends() {
+    use ::bytes::TryGetError;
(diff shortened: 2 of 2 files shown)


fallenmi (approved):
I used Codex to independently review exact base 76c0fbb54ed4336caf9d2311658a2f4a5627c21d against exact head 43284a0a5a709c7c299b1d2c15136e24f9b71031.

- A base-only oracle comparing fallible and infallible signed reads across widths 0 through 8 deterministically failed at width 1: try_get_int returned 255 where get_int returned -1. The short-read error and cursor-preservation control stayed green.
- The same big-endian and little-endian width matrix passes on the head, including width 0 and full-width 8. A short read still returns TryGetError { requested: 3, available: 2 } without advancing the buffer.
- The exact head passed the 876-case test_buf target plus the two independent oracle tests.
- The two-file diff applied cleanly to current master d5c8ad3227afe459c09f1d0d85455abf00f0381a. There, cargo test --all-features passed, including 876 Buf cases and 248 doctests; cargo fmt --all --check and git diff --check were clean.
- All 18 live checks are successful across stable platforms, minrust, cross targets, no-std, loom, tsan, miri, rustfmt, and documentation.

Approved for exact head 43284a0a5a709c7c299b1d2c15136e24f9b71031.

teddytennant:
Anything I can do to move this along? Happy to rebase if you want CI run against current master.

Darksonn (approved):
Thanks.

PR #842, opened 2026-07-14
Release shared reference on zero truncate

### LINKED ISSUE #840: `Bytes::clear()` and `Bytes::truncate(0)` do not decrement refcount
Since `Bytes::clear()` sets the length to zero and the length cannot be increased again, I assumed it must decrement the internal refcount so the parent buffer can be reused or freed.

However, looking through the code, I was surprised to learn that this wasn't the case; `Bytes::truncate(0)` (called by `Bytes::clear()`) [merely sets the length to zero without touching the refcount](https://github.com/tokio-rs/bytes/blob/76c0fbb54ed4336caf9d2311658a2f4a5627c21d/src/bytes.rs#L579).

Thus an empty `Bytes` can still retain a reference to the parent buffer and prevent it from being reused or freed, potentially resulting in leaks or unnecessary allocations.

I can't find any code comments or existing issues that explain whether this is the intended behavior or not. The only use-case I could see is that it allows the buffer to be upgraded to `BytesMut` if it remains the only handle, but this is not documented as an intentional side-effect.

Meanwhile, `Bytes::split_off(0)` [sets `self` to a new empty instance with the same pointer](https://github.com/tokio-rs/bytes/blob/76c0fbb54ed4336caf9d2311658a2f4a5627c21d/src/bytes.rs#L477-L479) to avoid touching the refcount, which results in _decrementing_ the refcount if one then drops the returned instance. 

However, this means that `Bytes::split_off(0)` doesn't allow `self` to be upgraded since it replaces the vtable with `STATIC_VTABLE`. This seems like an intentional side-effect, which only makes me more confused as to why `Bytes::clear()` doesn't have the same side-effect.

JSap0914:
Fixes #840.

When `Bytes::truncate(0)` replaces a shared handle with the zero-length representation, release the old handle so its shared reference is decremented. `Bytes::clear()` delegates to `truncate(0)`.

Verification:
- Focused: `cargo test --test test_bytes truncate_to_zero_releases_shared_reference -- --exact`
- Full: `cargo test`

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -567,7 +567,9 @@ impl Bytes {
     /// ```
     #[inline]
     pub fn truncate(&mut self, len: usize) {
-        if len < self.len {
+        if len == 0 {
+            drop(mem::replace(self, Bytes::new_empty_with_ptr(self.ptr)));
+        } else if len < self.len {
             // The Vec "promotable" vtables do not store the capacity,
             // so we cannot truncate while using this repr. We *have* to
             // promote using `split_off` so the capacity can be stored.
--- a/tests/test_bytes.rs
+++ b/tests/test_bytes.rs
@@ -340,6 +340,31 @@ fn truncate() {
 }
 
 #[test]
+fn truncate_to_zero_releases_shared_reference() {
+    let mut bytes = BytesMut::from(&b"hello"[..]);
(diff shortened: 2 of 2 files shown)


Darksonn (approved):
Thanks.

PR #844, opened 2026-07-24
bytes: add try_unwrap_or_clone method to Bytes for zero-alloc unwrap fast-path

Aditya-9-6:
### Summary
This PR adds `try_unwrap_or_clone(self) -> BytesMut` to `Bytes` in `src/bytes.rs`.

### Rationale & Performance Benefit
When converting a `Bytes` handle back into a mutable `BytesMut` buffer, `try_into_mut()` returns an `Err(Bytes)` if the handle is not unique. Callers currently have to manually match on `Ok` / `Err` and perform slice cloning.

Adding `try_unwrap_or_clone`:
- Provides a clean zero-allocation fast-path when the `Bytes` handle is unique.
- Clones underlying bytes into a new `BytesMut` buffer only when shared.
- Streamlines buffer conversion pipelines in high-throughput network applications.

### DIFF
--- a/src/bytes.rs
+++ b/src/bytes.rs
@@ -625,6 +625,28 @@ impl Bytes {
         }
     }
 
+    /// Consumes `self` and returns a `BytesMut`.
+    ///
+    /// If `self` is unique for the entire original buffer, this will return a `BytesMut`
+    /// referencing the same memory buffer without allocating. Otherwise, it will clone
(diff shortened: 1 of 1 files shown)


Darksonn:
I don't think we should add this method. The end user can easily do this manually.

By the way, the name is confusing. It says `_or_clone`, but normally clone in this context means incrementing the refcount, but this actually copies the data instead.

PR #846, opened 2026-08-29
Add defmt support behind defmt feature

SimonIT:


seanmonstar:
We are very unlikely to add in new public cargo dependencies to this crate.

SimonIT:
That's very sad, as defmt is the de facto standard logging crate in the embedded space. Bytes as no_std crate is also in the same category

PR #849, opened 2026-09-09
Optimize Extend when kind is a Vec

### LINKED ISSUE #848: Optimize Extend trait to have BytesMut::extend(&mut self, &[u8) compile to a `memcpy` when possible
I am currently working on a library that copies a data slice received from a QUIC library into a passed in buffer. I didn't want to force a specific buffer type so I have the passed in buffer implement `for<'a> Extend<&'a u8>`. Currently if a `Vec` is passed in as the buffer it compiles down to a `memcpy`. This makes sense because [`Vec`'s docs](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Extend%3C%26T%3E-for-Vec%3CT,+A%3E) state:

> This implementation is specialized for slice iterators, where it uses [copy_from_slice](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) to append the entire slice at once.

Meanwhile `BytesMut` copies data byte-by-byte, even when compiled in release mode.

Here's a [playground link](https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=8aed4e76bdf6c370fe38c59866aab854) with the following source code to verify that `BytesMut` copies data byte-by-byte:

```rust
use bytes::{Bytes, BytesMut};

#[inline(always)]
fn read_into_buf(bytes_from_network: &[u8], buf: &mut impl for<'a> Extend<&'a u8>) -> () {
    buf.extend(bytes_from_network);
}

#[unsafe(no_mangle)]
fn with_vec(bytes_from_network: &[u8]) {
    let mut into = Vec::new();
    read_into_buf(bytes_from_network, &mut into);
}

#[unsafe(no_mangle)]
fn with_bytes_mut(bytes_from_network: &[u8]) {
    let mut into = BytesMut::new();
    read_into_buf(bytes_from_network, &mut into);
}
```

The BytesMut Extend implementation currently has this as a TODO comment:
https://github.com/tokio-rs/bytes/blob/7930d93d583757e922014312d4aa3e58a415131f/src/bytes_mut.rs#L1494-L1498

I was wondering if there was a reason this optimization has not been implemented yet. It would allow `extend` calls, like the one in my library, to turn into `memcpy` instead of copying data byte-by-byte.

zphrs:
Fixes #848


zphrs:
Forgot to run tests and my code had a bug; I'm working on a fix currently.

zphrs:
I forgot to update length after extending because I based my code on the reserve() implementation. It's fixed now.

PR #857, opened 2026-09-22
Implement From<Arc<[u8]>> for Bytes

### LINKED ISSUE #685: Confusing documentation around `Arc<[u8]>` compatibility
The latest documentation of the crate reads:

```
For Bytes implementations which point to a reference counted shared storage (e.g. an Arc<[u8]>),
sharing will be implemented by increasing the reference count.
```
Reference: https://docs.rs/bytes/latest/bytes/struct.Bytes.html#sharing

This lead me to assume that Bytes should provide a `from` conversion function from an `Arc<[u8]>` to a `Bytes` instance where the backing storage is an `Arc`. However, it doesn't appear that this `from` function exists.

I don't think it's a good idea for documentation to reference imaginary scenarios that aren't implemented by the library as it sets up false expectations with regard to how the library works and what features it provides.

sb123sb123:
Fixes https://github.com/tokio-rs/bytes/issues/685

## Problem

The `Bytes` documentation describes reference-counted backing storage such as `Arc<[u8]>`, but `Bytes` had no `From<Arc<[u8]>>` implementation. The documented conversion therefore failed to compile.

## Cause

`Bytes::from_owner` already supports owning any `AsRef<[u8]> + Send + 'static` value, but the standard `Arc<[u8]>` conversion was missing from the `From` implementations.

## Fix

Add a `target_has_atomic = "ptr"`-gated `From<alloc::sync::Arc<[u8]>> for Bytes` implementation that delegates to `Bytes::from_owner`. Add a regression test that drops the original `Arc`, verifies the bytes remain usable, and checks that the data pointer is preserved.

## Tests

- `cargo fmt --all -- --check`
- `cargo test --test test_bytes from_arc -- --nocapture`
- `cargo test --all-features`
- `cargo check --no-default-features`
- `cargo clippy --all-features --lib` (passes with existing warnings)
- `git diff --check`
- Independent pre-fix reproduction failed with the missing `From<Arc<[u8]>>` trait implementation; the same reproduction compiles after the fix.

## Limitations

- The checks ran on Windows. Linux/macOS, cross-target, no-atomic-target, nightly/Miri, and the full CI platform matrix were not run locally.
- `cargo clippy --all-features --all-targets -- -D warnings` remains red on clean HEAD because of pre-existing diagnostics in unrelated code; the same baseline failure was reproduced before this change.
- The conversion is gated on pointer-width atomics because `alloc::sync::Arc` is unavailable on targets without atomic pointers.

## AI assistance

This pull request was prepared with AI assistance. AI was used for repository and issue triage, reproduction, implementation, and test planning; the behavior, patch, and test results were independently checked on the remote host.


tests/test_bytes.rs
```diff
@@ -67,6 +67,17 @@ fn from_slice() {
     assert_eq!(Vec::from(&b"abcdefgh"[..]), a);
 }
 
+#[test]
```
martin-g, line 70:
```suggestion
#[cfg(target_has_atomic = "ptr")]
#[test]
```

